PR #10, opened 2021-02-18
[C-Api] fix res leak in timeout mode

jaeyun-jung:
In single-shot timeout case, the flag to release output handle may reset when setting short timeout value.
To prevent this case, add output handle to del-list and change params to set input/output handle in invoke thread.


gichan-jang (approved):
LGTM!

myungjoo:
PTAL: @leemgs @again4you

again4you (approved):
Nice catch!

PR #17, opened 2021-03-08
[Test] tensor name in tf-lite tc

jaeyun-jung:
Ignore tensor name comparison in tf-lite testcase.
NNStreamer tf-lite plugin will set tensor name.


gichan-jang (approved):
LGTM

PR #21, opened 2021-03-22
[Packaging] api devel requires nnstreamer-devel @open sesame 03/22 16:01

zhoonit:
This patch fixes api devel to require nnstreamer-devel so that devel
does not depending on nnstreamer which would conflicts with
`nnstreamer-test-devel`

**Self evaluation:**

### DIFF
--- a/packaging/machine-learning-api.spec
+++ b/packaging/machine-learning-api.spec
@@ -144,8 +144,8 @@ You can construct a data stream pipeline with neural networks easily.
 %package devel
 Summary:	Tizen Native API Devel Kit for NNStreamer
 Group:		Machine Learning/ML Framework
-Requires:	capi-machine-learning-inference = %{version}-%{release}
-Requires:	capi-machine-learning-common-devel
+Requires:	nnstreamer-devel
+Requires:	capi-machine-learning-common-devel = %{version}-%{release}
 %description devel
 Developmental kit for Tizen Native NNStreamer API.
 


packaging/machine-learning-api.spec
```diff
@@ -242,6 +242,7 @@ DESTDIR=%{buildroot} ninja -C build %{?_smp_mflags} install
 %files devel
 %{_includedir}/nnstreamer/nnstreamer.h
 %{_includedir}/nnstreamer/nnstreamer-single.h
+%{_libdir}/libcapi-nnstreamer.so.*
```
again4you, line 245: What if developers try to install both `capi-machine-learning-inference` and `capi-machine-learning-inference-devel`, the conflict occurs. Is this your intent?

zhoonit: `capi-machine-learning-inference-devel` package breaks if `capi-machine-learning-inference` is installed, and it will be so hard to find the reason why the devel package is not working for the developer.

So I think it's okay to let it conflict.

wooksong, line 245: In general, a **devel** package does include ```.so``` only rather than all of the ```.so.*```.

zhoonit: I checked that the file and it was a symlinked to `.so.*`. should we make this as a standalone file?

wooksong: > I checked that the file and it was a symlinked to .so.*

Yes, that is what exactly .so file is.

wooksong: And the same line appears the one at 240.

zhoonit: Understood and deleted. Thank you for giving me the opportunity to learn 👍

myungjoo:
Do not fix this by not requiring nnstreamer-configuration, but by nnstreamer-test-devel mimic nnstrreamer-configuration.

Hint 1: Requires, Recommends, Provides, "concept: Meta Package", "concept: Capability of RPM". 
Hint 2: You may need to replace Conflicts with other dependencies in nnstreamer.spec.

packaging/machine-learning-api.spec
zhoonit: This is reverted as suggested right below.

zhoonit:
Consulted @myungjoo about the matter.

I was recommended to add a virtual capabilities of configuration have a `recommends` to actual conf package when building `nnstreamer` and in this case building this package will have no change, so closing this PR.

PR #22, opened 2021-03-24
[Log] divide a log message

helloahn:
Since the length of the pipeline string is too long, the error message cannot be checked.
For easier debugging, this log message should be divided.

Signed-off-by: Hyoung Joo Ahn <hello.ahn@samsung.com>


anyj0527 (approved):
LGTM 👍

myungjoo (approved):

PR #26, opened 2021-03-29
Fix unit test script

myungjoo:
In main routine, return cannot be used.
Return values by calling exit instead.


packaging/run_unittests.sh
```diff
@@ -18,7 +18,7 @@ run_entry() {
 if [ -f "$1" ]; then
     echo $1
     run_entry $1
-    return $?
+    exit $?
```
jaeyun-jung, line 21: this change is required in nntrainer repo :)

jaeyun-jung (approved):

PR #30, opened 2021-04-06
[Test] change property name

jaeyun-jung:
change property name sync mode, nnstreamer will change sync mode and option.


again4you (approved):
Nice catch!

PR #39, opened 2021-04-23
[C-Api] handle flexible tensor

jaeyun-jung:
In flexible tensor stream, sink and src handle cannot set exact tensor info from pad caps.
To handle flex-tensor, ignore tensor info while configuring the element handle.

### DIFF
--- a/c/include/nnstreamer-capi-private.h
+++ b/c/include/nnstreamer-capi-private.h
@@ -256,6 +256,7 @@ typedef struct _ml_pipeline_element {
 
   GMutex lock; /**< Lock for internal values */
   gboolean is_media_stream;
+  gboolean is_flexible_tensor;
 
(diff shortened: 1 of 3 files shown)


jaeyun-jung:
Requires nnstreamer PR https://github.com/nnstreamer/nnstreamer/pull/3212

myungjoo (approved):
LGTM

myungjoo:
Do you need additional modifications for the stream type consolidation? https://github.com/nnstreamer/nnstreamer/issues/3351

jaeyun-jung:
@myungjoo we need changing mime-type in testcases later. IMO, updating API code is unnecessary after implementing https://github.com/nnstreamer/nnstreamer/issues/3351.

myungjoo (approved):

PR #40, opened 2021-04-26
[Android/Pipeline] clone data from src element

jaeyun-jung:
We cannot guarantee the data object from java is available while running the pipeline.
Set clone when pushing data to src element.


anyj0527 (approved):
I verified this PR with local ml-api test and Android App.
LGTM👍👍

again4you (approved):
LGTM!

PR #42, opened 2021-04-29
Thread safety for ml_tensors_info APIs

myungjoo:
Fixes #nnstreamer/2600

https://github.com/nnstreamer/nnstreamer/issues/2600

CC: @kparichay 

ps. This updates internal behaviors. No need for ACR.

### DIFF
--- a/c/src/nnstreamer-capi-util.c
+++ b/c/src/nnstreamer-capi-util.c
@@ -57,6 +57,7 @@ ml_tensors_info_create (ml_tensors_info_h * info)
     ml_loge ("Failed to allocate the tensors info handle.");
     return ML_ERROR_OUT_OF_MEMORY;
   }
+  g_mutex_init (&tensors_info->lock);
 
   /* init tensors info struct */
   return ml_tensors_info_initialize (tensors_info);
@@ -81,6 +82,7 @@ ml_tensors_info_create_from_gst (ml_tensors_info_h * ml_info,
     ml_loge ("Failed to allocate the tensors info handle.");
     return ML_ERROR_OUT_OF_MEMORY;
   }
+  g_mutex_init (&tensors_info->lock);
 
   /* init and copy tensors info from gst struct */
   ml_tensors_info_initialize (tensors_info);
@@ -103,7 +105,11 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
   if (!tensors_info)
     return ML_ERROR_INVALID_PARAMETER;
 
+  G_LOCK_UNLESS_NOLOCK (*tensors_info);
+
   ml_tensors_info_free (tensors_info);
+  G_UNLOCK_UNLESS_NOLOCK (*tensors_info);
+  g_mutex_clear (&tensors_info->lock);
   g_free (tensors_info);
 
   return ML_ERROR_NONE;
@@ -136,6 +142,7 @@ ml_tensors_info_initialize (ml_tensors_info_s * info)
 
 /**
  * @brief Validates the given tensor info is valid.
+ * @note info should be locked by caller if nolock == 0.
  */
 static gboolean
 ml_tensor_info_validate (const ml_tensor_info_s * info)
@@ -187,6 +194,7 @@ ml_tensors_info_validate (const ml_tensors_info_h info, bool *valid)
 {
   ml_tensors_info_s *tensors_info;
   guint i;
+  int ret = ML_ERROR_NONE;
 
   check_feature_state ();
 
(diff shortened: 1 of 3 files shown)


c/src/nnstreamer-capi-util.c
```diff
@@ -261,6 +294,8 @@ ml_tensors_info_set_count (ml_tensors_info_h info, unsigned int count)
     return ML_ERROR_INVALID_PARAMETER;
 
   tensors_info = (ml_tensors_info_s *) info;
+
+  /* This is atomic. No need for locks */
```
kparichay, line 298: I am not sure if this is an atomic operation, as this assignment can be broken down into more than 1 instruction by the compiler.

c/src/nnstreamer-capi-util.c
```diff
@@ -198,6 +225,7 @@ ml_tensors_info_validate (const ml_tensors_info_h info, bool *valid)
   if (!tensors_info || tensors_info->num_tensors < 1)
```
kparichay, line 225: best to split this check into two parts and move the latter inside the lock.

kparichay (commented):
Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).

c/src/nnstreamer-capi-util.c
myungjoo: 32bit (unsigned int) assignment cannot be more than 1 instruction. We may incur multiple instructions for redirections (```tensors_info->num_tensors```) and load (```LOAD Rx count```).
However, these does not affect the atomicity of assignment; they are allowed to be executed out of the critical section.
We cannot keep it serialized with destruction; thus, redirection is not protected anyway.
The location of ```count``` is never affected by the serialization problem; thus, atomicity is not affected.

Thus, we simply need to protect ```MOV Ra Rb``` instruction here.

c/src/nnstreamer-capi-util.c
myungjoo: split done.

myungjoo:
> Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
> Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).

added.

jaeyun-jung (approved):

kparichay:
> > Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
> > Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).
> 
> added.

Can you please elaborate on how do you handle this? I can't see any `changes in ml_pipeline_custom_easy_filter_register()`.

myungjoo:
> > > Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
> > > Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).
> > 
> > 
> > added.
> 
> Can you please elaborate on how do you handle this? I can't see any `changes in ml_pipeline_custom_easy_filter_register()`.

ah.. only clone was added. I'll add easy_filter soon.

myungjoo:
> > > > Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
> > > > Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).
> > > 
> > > 
> > > added.
> > 
> > 
> > Can you please elaborate on how do you handle this? I can't see any `changes in ml_pipeline_custom_easy_filter_register()`.
> 
> ah.. only clone was added. I'll add easy_filter soon.

For ml_pipeline_custom_easy_filter_register(), it becomes safe if we don't access in or out after ml_tensors_info_clone; thus, I'll modify parameter usages of ml_tensors_info_copy_from_ml.

kparichay:
> > > > > Please add locks to `ml_tensors_info_clone()` as well, as it is an exposed API.
> > > > > Also, for `ml_pipeline_custom_easy_filter_register()`, although each access to `ml_tensors_info_h` is locked, but are locked individually, and results in TOCTOU (tensors_info validity is checked and then later copied, but lock is released in between the checking and copying).
> > > > 
> > > > 
> > > > added.
> > > 
> > > 
> > > Can you please elaborate on how do you handle this? I can't see any `changes in ml_pipeline_custom_easy_filter_register()`.
> > 
> > 
> > ah.. only clone was added. I'll add easy_filter soon.
> 
> For ml_pipeline_custom_easy_filter_register(), it becomes safe if we don't access in or out after ml_tensors_info_clone; thus, I'll modify parameter usages of ml_tensors_info_copy_from_ml.

I think there is another issue of `tensors_info_h` being lock free between `ml_tensors_info_validate` and `ml_tensors_info_clone`.  As `ml_tensors_info_clone` does not check `ml_tensors_info_validate` on its own, it possible for a tensor to pass the validity check, but become non-valid before it could be cloned. Further, there is no check on result of the clone which I think is because we assume that all the tensors are valid, however that is not the case anymore because of the error of Time Of Check Time Of Use. I will make a separate PR to resolve this.

PR #46, opened 2021-05-06
[Spec] Add the provides section to capi-machine-learning-inference package @open sesame 5/06 19:40

again4you:
Since rpmbuild explicitly exclude the name from the library rpm,
libcapi-nnstreamer.so is not provided as below.

```bash
$ rpm -qp --provides capi-machine-learning-inference-1.7.2-0.armv7l.rpm
capi-machine-learning-inference = 1.7.2-0
capi-machine-learning-inference(armv7l-32) = 1.7.2-0
libcapi-nnstreamer.so.1
```

This patch adds the provides section to capi-machine-learning-inference
package so that capi-machine-learning-inference provides
libcapi-nnstreamer.so.

### DIFF
--- a/packaging/machine-learning-api.spec
+++ b/packaging/machine-learning-api.spec
@@ -64,6 +64,7 @@ Source1001:	capi-machine-learning-inference.manifest
 
 ## Define build requirements ##
 Requires:	nnstreamer
+Provides:	libcapi-nnstreamer.so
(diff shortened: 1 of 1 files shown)


myungjoo (approved):
This is ...  workaround, not a recommended fix. But, for now, it appears to be the only option available.

jaeyun-jung (approved):

PR #52, opened 2021-05-11
Thread safety for ml_tensors_data APIs @open sesame 5/27 17:10

myungjoo:
Fixes #nnstreamer/2600
Note that locally created-and-freed ml-tensors-data
are not required to be protected.
    
### DIFF
--- a/c/src/nnstreamer-capi-pipeline.c
+++ b/c/src/nnstreamer-capi-pipeline.c
@@ -8,7 +8,7 @@
  *        This allows to construct and control NNStreamer pipelines.
  * @see	https://github.com/nnstreamer/nnstreamer
  * @author MyungJoo Ham <myungjoo.ham@samsung.com>
- * @bug No known bugs except for NYI items
+ * @bug Thread safety for ml_tensors_data should be addressed.
  */
 
 #include <string.h>
@@ -293,6 +293,7 @@ cb_sink_event (GstElement * e, GstBuffer * b, gpointer user_data)
     ml_loge ("Failed to allocate memory for tensors data in sink callback.");
     return;
   }
+  g_mutex_init (&data->lock);
 
   data->num_tensors = num_mems;
   for (i = 0; i < num_mems; i++) {
@@ -399,6 +400,7 @@ error:
   }
 
   if (data) {
+    g_mutex_clear (&data->lock);
     g_free (data);
     data = NULL;
   }
@@ -1426,12 +1428,13 @@ ml_pipeline_src_input_data (ml_pipeline_src_h h, ml_tensors_data_h data,
     ret = ML_ERROR_INVALID_PARAMETER;
     goto unlock_return;
   }
+  G_LOCK_UNLESS_NOLOCK (*_data);
 
   if (_data->num_tensors < 1 || _data->num_tensors > ML_TENSOR_SIZE_LIMIT) {
     ml_loge ("The tensor size is invalid. It should be 1 ~ %u; where it is %u",
         ML_TENSOR_SIZE_LIMIT, _data->num_tensors);
     ret = ML_ERROR_INVALID_PARAMETER;
-    goto destroy_data;
+    goto dont_destroy_data;
   }
 
   ret = ml_pipeline_src_parse_tensors_info (elem);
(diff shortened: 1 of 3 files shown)


c/src/nnstreamer-capi-pipeline.c
```diff
@@ -293,6 +293,7 @@ cb_sink_event (GstElement * e, GstBuffer * b, gpointer user_data)
     ml_loge ("Failed to allocate memory for tensors data in sink callback.");
     return;
   }
+  g_mutex_init (&data->lock);
```
jaeyun-jung, line 296: need clear at line 402

myungjoo: applied.

c/src/nnstreamer-capi-pipeline.c
```diff
@@ -1480,11 +1483,17 @@ ml_pipeline_src_input_data (ml_pipeline_src_h h, ml_tensors_data_h data,
     /** @todo Verify that gst_buffer_append lists tensors/gstmem in the correct order */
   }
 
+  /* Unlock if it's not auto-free. We do not know when it'll be freed. */
```
jaeyun-jung, line 1486: line 1486 ~ 1515
1. deadlock case if data is invalid 
2. If given policy is do-not-free case, buffers are wrapped and pushed into pipeline. So I think this unlock code (1486 ~1488) is unnecessary.

how about adding unlock in line 1511?
```
destroy_data:
  if (_data != NULL) {
    G_UNLOCK_UNLESS_NOLOCK (*_data);

    /* Free data handle */
    if (policy == ML_PIPELINE_BUF_POLICY_AUTO_FREE)
      ml_tensors_data_destroy (data);
  }

  handle_exit (h);
}
```

myungjoo:
It hangs. It may have a deadlock.

c/src/nnstreamer-capi-pipeline.c
myungjoo: > 1. deadlock case if data is invalid

Thanks for finding out the deadlock!

> 2. If given policy is do-not-free case, buffers are wrapped and pushed into pipeline. So I think this unlock code (1486 ~1488) is unnecessary.

If we unlock at destroy_data, the user's thread may call ```free(_data)``` before G_UNLOCK_UNLESS_NOLOCK. Thus, we need to UNLOCK before calling push_buffer.

c/src/nnstreamer-capi-pipeline.c
```diff
@@ -1426,12 +1428,14 @@ ml_pipeline_src_input_data (ml_pipeline_src_h h, ml_tensors_data_h data,
...
     goto unlock_return;
   }
+  G_LOCK_UNLESS_NOLOCK (*_data);
 
   if (_data->num_tensors < 1 || _data->num_tensors > ML_TENSOR_SIZE_LIMIT) {
     ml_loge ("The tensor size is invalid. It should be 1 ~ %u; where it is %u",
         ML_TENSOR_SIZE_LIMIT, _data->num_tensors);
     ret = ML_ERROR_INVALID_PARAMETER;
-    goto destroy_data;
+    G_UNLOCK_UNLESS_NOLOCK (*_data);
```
jaeyun-jung, line 1437: line 1446, 1456, 1468
requires unlock if policy == ML_PIPELINE_BUF_POLICY_DO_NOT_FREE

myungjoo: Thanks. actually, they should not call destroy-data (another buf found!) if push-buffer is not called. Fix is coming.

jaeyun-jung (approved):

c/src/nnstreamer-capi-pipeline.c
```diff
@@ -1498,11 +1507,15 @@ ml_pipeline_src_input_data (ml_pipeline_src_h h, ml_tensors_data_h data,
     ret = ML_ERROR_STREAMS_PIPE;
   }
 
-destroy_data:
   if (_data != NULL && policy == ML_PIPELINE_BUF_POLICY_AUTO_FREE) {
     /* Free data handle */
+    G_UNLOCK_UNLESS_NOLOCK (*_data);
     ml_tensors_data_destroy (data);
   }
+  goto unlock_return;
```
gichan-jang, line 1515: I cannot find `unlock_return` label.

anyj0527: It's in macro `handle_exit`

PR #58, opened 2021-06-04
[CAPI/ml-common] Add require to base-common

### LINKED ISSUE #57: Capi-ml-common is depending on capi-base-common
in the `capi-ml-common` pc file we should be requiring `capi-base-common` and our meson should not directly require `capi-base-common`

zhoonit:
- [CAPI/ml-common] Add require to base-common

```
When a developer tries to add dependency to `capi-ml-common` they have
to care about adding dependency to `capi-base-common` as well because
`capi-ml-common` is actually depending upon `capi-base-common`

This patch resolves the issue by adding require to `capi-base-common`
in the pc file.

resolves #57

**Self evaluation:**

```

resolves #57

### DIFF
--- a/c/meson.build
+++ b/c/meson.build
@@ -40,7 +40,6 @@ if (get_option('enable-tizen'))
   message('C-API is in Tizen mode')
 
   tizen_deps = [
-    dependency('capi-base-common'),
     dependency('capi-system-info'),
     dependency('dlog')
   ]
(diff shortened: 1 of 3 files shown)


myungjoo (approved):
LGTM

c/meson.build
```diff
@@ -90,9 +89,14 @@ configure_file(input: 'capi-ml-inference.pc.in', output: 'capi-ml-inference.pc',
   configuration: api_conf
 )
 
+ml_api_conf = configuration_data()
```
jaeyun-jung, line 92: recommend to move this and change .pc also (line 89 configuration: api_conf > ml_api_conf)

zhoonit: Thank you for the suggestion! updated as per offline discussion

zhoonit: Thank you for the suggestion! Changed as per offline discussion :)

jaeyun-jung (approved):

PR #61, opened 2021-06-09
[Flatbuf] Change version check of the flatc @open sesame 06/09 17:21

gichan-jang:
Change version check of the flatc from error to warning.
Even if the flatbuffers compiler has different versions, compatibility can be maintained.
Tested with
 - System flatc version: 2.0.0
 - Android flatc version: 1.12.0

Self evaluation:
1. Build test: [ *]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [* ]Skipped


jaeyun-jung (approved):

again4you (approved):
If there is no potential problem, it's OK.

PR #66, opened 2021-06-21
[Android] nnfw feature

jaeyun-jung:
Change nnfw feature definition in android build, same to c-api define.


anyj0527 (approved):
LGTM 👍👍

PR #68, opened 2021-06-23
[Java] prepare ubuntu build @open sesame 06/23 19:10

jaeyun-jung:
Prepare Java API on ubuntu, add feature for android and remove android annotation.

Signed-off-by: Jaeyun <jy1210.jung@samsung.com>

PR #70, opened 2021-06-23
[Java] build JNI wrapper

jaeyun-jung:
Build JNI wrapper using meson.
The source for JNI wrapper is same with Android native code.


jaeyun-jung:
Be careful, when building JNI wrapper using meson, PR #68 is required to prevent build error.

myungjoo (approved):
LGTM

myungjoo:
We are  going to provide both JAVA-API and Android-JAVA-API soon!

PR #71, opened 2021-06-23
[Android/Build] check target abi

jaeyun-jung:
Check target abi when build android library, and set dir name to get gstreamer prebuilt libs.


myungjoo (approved):
LGTM

PR #72, opened 2021-06-24
[Build/Java] add script to build jar

jaeyun-jung:
Add new script to build java library.
To prevent build error, mark prefix in java files and change it while building the library.


myungjoo (approved):
LGTM if tested in both platforms

anyj0527 (approved):
LGTM 👍

PR #75, opened 2021-06-30
[JNI] remove unnecessary feature in native

jaeyun-jung:
We should call subplugin registration only on Android.
Remove unnecessary feature in native code.


again4you (approved):
LGTM!

PR #79, opened 2021-07-19
[C-Api] remove warning while checking framework

jaeyun-jung:
When checking fw with model files is failed, it returns null and this will print warning message even if this works properly.
Update macro to not print this message. (requested from VD)


gichan-jang (approved):
LGTM

again4you (approved):
LGTM!

PR #80, opened 2021-07-19
[C-Api] framework priority from conf

jaeyun-jung:
Load framework priority from nnstreamer conf file.


again4you (approved):
LGTM!

PR #86, opened 2021-08-06
Unittest: data type consistency with nnstreamer/gstreamer @open sesame 8/13 11:56 

myungjoo:
There are type re-definitions in C ML-API, which
should inherit values from nnstreamer and gstreamer.
Add unittests that ensure consistency.

Fixes https://github.com/nnstreamer/nnstreamer/issues/2587
CC: @kparichay 

### DIFF
--- a/tests/capi/unittest_capi_datatype_consistency.cc
+++ b/tests/capi/unittest_capi_datatype_consistency.cc
@@ -1,0 +1,122 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/**
+ * @file        unittest_capi_datatype_consistency.cc
+ * @date        06 Aug 2021
+ * @brief       Unit test for data type consistency with frameworks
+ * @see         https://github.com/nnstreamer/api
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs
+ */
+
+/* NNStreamer Side */
+#include <tensor_typedef.h>
+
+/* ML API Side */
+#include <nnstreamer.h>
+#include <nnstreamer-capi-private.h>
+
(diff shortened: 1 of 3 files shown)


kparichay (approved):
Great Job. The tests match the enums and size constraints.
For nnstreamer/nnstreamer#2587, it would be even better if we can have two more tests to ensure that the pairs {ml_tensor_info_s, GstTensorInfo} and {ml_tensors_data_s, GstTensorMemory} can be type-casted interchangeably (removing the minor overhead of creating these data structures between ML-API and NNStreamer). But this will limit the flexibilty of ml_tensor_info/data_s for future modifications (IMO, these won't need modifications but just noting this as a constraint).

myungjoo:
Need to pass https://dashboard.tizen.org/index.code?sr=submit/tizen/20210809.021120 to make this buildable.

tests/capi/unittest_capi_datatype_consistency.cc
```diff
@@ -0,0 +1,119 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/**
+ * @file        unittest_capi_datatype_consistency.cc
+ * @date        06 Aug 2021
+ * @brief       Unit test for data type consistency with frameworks
+ * @see         https://github.com/nnstreamer/api
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs
+ */
+
```
jaeyun-jung, line 10: need `#include <gtest/gtest.h>` :)

myungjoo: Thanks! added.

PR #89, opened 2021-08-18
[Android/Build] remove release flag

jaeyun-jung:
Remove release flag and related source.
Uploading library to jcenter is not available anymore.


anyj0527 (approved):
We may consider releasing on Maven later... LGTM 👍

PR #94, opened 2021-09-24
[Single] Don't close single handle during invoke process

gichan-jang:
Don't close single handle during invoke process.
If single handle is released during invoke prcoess, an interpreter may
access invalid memory.


c/src/nnstreamer-capi-single.c
```diff
@@ -841,8 +844,12 @@ int
 ml_single_close (ml_single_h single)
 {
   ml_single *single_h;
+  GMutex invoking_lock;
+  gboolean invoking;
+
```
again4you, line 849: Unnecessary newline?!

gichan-jang: Removed.

jaeyun-jung (approved):

PR #96, opened 2021-10-19
Refactoring C-API module relations.

myungjoo:
ML C-API interfaces and implementations are being refactored for better relations.

ml-common will be no longer depending on nnstreamer.

Then, in the future, nntrainer will be no longer depending on nnstreamer except for its nnstreamer-nntrainer subplugin.


myungjoo:
Incoming further refactoring.. do not merge.

PR #108, opened 2021-11-16
[Build/CodeClean] std headers

jaeyun-jung:
Code clean, it is recommended to not use stddef for NDK build, use stdlib instead.


zhoonit (approved):
LGTM!!

myungjoo (approved):

PR #111, opened 2021-11-22
[fix/trivial] Fix unreachable or unnecessary codes

anyj0527:
- Remove unreachable code
- Remove unnecessary null check

REF:
https://analysishub.sec.samsung.net/service/analyses/985737
https://analysishub.sec.samsung.net/service/analyses/985799


again4you (approved):
LGTM!

PR #126, opened 2022-01-04
[C/Inference] Add check routine for the product-specific restricted elements

again4you:
This patch newly adds the check routine for the product-specific restricted
elements. If a product-specific configuration feature is enabled, ML API
will check the restricted elements in that file. To enable the
product-specific configuration, the below condition should be matched.

* 'enable_product_conf' in 'common' group should be True.
* 'path' in 'product-config' should be set and the ini file should exist.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped

### Related issue
* https://github.com/nnstreamer/nnstreamer/issues/3629

### DIFF
--- a/c/src/ml-api-inference-internal.c
+++ b/c/src/ml-api-inference-internal.c
@@ -298,6 +298,44 @@ __ml_validate_model_file (const char *const *model,
 }
 
 /**
+ * @brief Internal helper function to read the string from specific ini file.
+ */
+static gchar *
+__ml_get_string_from_conf (const gchar * ini_path,
+    const gchar * group, const gchar * key)
+{
+  g_autofree gchar *value = NULL;
+  g_autoptr (GKeyFile) key_file = NULL;
+
+  if (!ini_path || !g_file_test (ini_path, G_FILE_TEST_EXISTS)) {
+    _ml_loge
+        ("The parameter, ini_path, is NULL. It should be a valid path of configuration file.");
+    return NULL;
+  }
+
+  if (!group) {
+    _ml_loge
+        ("The parameter, group, is NULL. It should be a valid string of group name.");
+    return NULL;
+  }
+
+  if (!key) {
+    _ml_loge
+        ("The parameter, key, is NULL. It should be a valid string of key.");
+    return NULL;
+  }
+
+  key_file = g_key_file_new ();
+  g_assert (key_file != NULL);
+
+  if (g_key_file_load_from_file (key_file, ini_path, G_KEY_FILE_NONE, NULL)) {
+    value = g_key_file_get_string (key_file, group, key, NULL);
+  }
+
+  return g_strdup (value);
+}
+
+/**
  * @brief Internal function to get the nnfw type.
  */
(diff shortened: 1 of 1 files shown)


c/src/ml-api-inference-internal.c
```diff
@@ -297,6 +297,44 @@ __ml_validate_model_file (const char *const *model,
...
+static gchar *
+__ml_get_string_from_conf (const gchar * ini_path,
+    const gchar * group, const gchar * key)
+{
+  g_autofree gchar *value = NULL;
+  g_autoptr (GKeyFile) key_file = NULL;
+
+  if (!ini_path || !g_file_test (ini_path, G_FILE_TEST_EXISTS)) {
+    _ml_loge
+        ("The parameter, ini_path, is NULL. It should be a valid path of configuration file.");
```
myungjoo, line 312: If it is not null and g_file_test fails, you need to show the given file path and say the file does not exist.

c/src/ml-api-inference-internal.c
```diff
@@ -297,6 +297,44 @@ __ml_validate_model_file (const char *const *model,
...
+ * @brief Internal helper function to read the string from specific ini file.
+ */
+static gchar *
+__ml_get_string_from_conf (const gchar * ini_path,
+    const gchar * group, const gchar * key)
+{
+  g_autofree gchar *value = NULL;
+  g_autoptr (GKeyFile) key_file = NULL;
+
+  if (!ini_path || !g_file_test (ini_path, G_FILE_TEST_EXISTS)) {
```
myungjoo, line 310: Refer to https://people.gnome.org/~ryanl/glib-docs/glib-File-Utilities.html#GFileTest

You may need to see if it's a REGULAR file or not as well.

c/src/ml-api-inference-internal.c
```diff
@@ -297,6 +297,44 @@ __ml_validate_model_file (const char *const *model,
...
+  }
+
+  if (!key) {
+    _ml_loge
+        ("The parameter, key, is NULL. It should be a valid string of key.");
+    return NULL;
+  }
+
+  key_file = g_key_file_new ();
+  g_assert (key_file != NULL);
```
myungjoo, line 329: Please do not add g_assert in main code; this is a library for an application and applications may want to keep running (with some error handling) with an error in our library.
(you may add assert in test codes.)

Please add a proper error handling code instead.

c/src/ml-api-inference-internal.c
```diff
@@ -297,6 +297,44 @@ __ml_validate_model_file (const char *const *model,
   return ML_ERROR_NONE;
 }
 
+/**
+ * @brief Internal helper function to read the string from specific ini file.
```
myungjoo, line 301: Please mention that the returned value should be freed by the caller in @return .

again4you:
Another patch is already merged for this issue so this PR is closed.

PR #131, opened 2022-01-24
[Package] Separate single-shot and pipeline API.

gichan-jang:
- `machine-learning-inference` is separated to `machine-learning-inference-single` and `machine-learning-inference-pipeline`.
 - Move `_ml_validate_model_file` to `ml-api-inference-single`

Signed-off-by: Gichan Jang <gichan2.jang@samsung.com>

PR #137, opened 2022-03-08
Fix libcapi-nnstremer is empty due to linker optimization

zhoonit:
## Commits to be reviewed in this PR

<details><summary>Fix libcapi-nnstremer is empty due to linker optimization</summary><br />

From #131, libcapi-nnstreamer.so became empty because
libcapi-nnstreamer.so does not use any symbol from
libcapi-nnstreamer-single, libcapi-nnstreamer-pipeline.so

This should be semantically resolvable by using 'link-whole' argument in
meson but it is not implemented for shared_library.

This patch address the issue by disabling `-bas-needed=false` in meson
and add -Wl,--as-needed to global linker argument.

See also: https://github.com/mesonbuild/meson/issues/3046.

**Self evaluation:**

</details>

### DIFF
--- a/c/meson.build
+++ b/c/meson.build
@@ -112,11 +112,11 @@ nns_capi_single_static_lib = static_library ('capi-nnstreamer-single',
   install_dir: api_install_libdir,
 )
 
-nns_capi_single_lib = nns_capi_single_shared_lib
+nns_capi_single_dep = declare_dependency(link_with: nns_capi_single_shared_lib)
 if get_option('default_library') == 'static'
-  nns_capi_single_lib = nns_capi_single_static_lib
+  nns_capi_single_dep = declare_dependency(link_with: nns_capi_single_static_lib)
 endif
-nns_capi_single_dep = declare_dependency(link_with: nns_capi_single_lib)
+
 
 # Pipeline API.
 nns_capi_pipeline_shared_lib = shared_library ('capi-nnstreamer-pipeline',
@@ -136,15 +136,15 @@ nns_capi_pipeline_static_lib = static_library ('capi-nnstreamer-pipeline',
   install_dir: api_install_libdir,
 )
 
-nns_capi_pipeline_lib = nns_capi_pipeline_shared_lib
+nns_capi_pipeline_dep = declare_dependency(link_with: nns_capi_pipeline_shared_lib)
 if get_option('default_library') == 'static'
-  nns_capi_pipeline_lib = nns_capi_pipeline_static_lib
+  nns_capi_pipeline_dep = declare_dependency(link_with: nns_capi_pipeline_static_lib)
 endif
-nns_capi_pipeline_dep = declare_dependency(link_with: nns_capi_pipeline_lib)
+
 
 # Single-shot and pipeline API.
 nns_capi_shared_lib = shared_library ('capi-nnstreamer',
-  link_with: [nns_capi_single_lib, nns_capi_pipeline_lib],
+  link_with: [nns_capi_single_shared_lib, nns_capi_pipeline_shared_lib],
   include_directories: nns_capi_include,
   install: true,
   install_dir: api_install_libdir,
@@ -152,7 +152,7 @@ nns_capi_shared_lib = shared_library ('capi-nnstreamer',
 )
 
 nns_capi_static_lib = static_library ('capi-nnstreamer',
-  link_with: [nns_capi_single_lib, nns_capi_pipeline_lib],
+  link_whole: [nns_capi_single_static_lib, nns_capi_pipeline_static_lib],
   include_directories: nns_capi_include,
   install: true,
   install_dir: api_install_libdir,
--- a/meson.build
+++ b/meson.build
@@ -3,12 +3,16 @@ project('ml-api', 'c', 'cpp',
   license: ['Apache-2.0'],
   meson_version: '>=0.50.0',
   default_options: [
+    'b_asneeded=false',
     'werror=true',
     'warning_level=1',
     'c_std=gnu89',
     'cpp_std=c++14'
   ]
 )
+
+add_global_link_arguments('-Wl,--no-as-needed', language: 'c')
+add_global_link_arguments('-Wl,--no-as-needed', language: 'cpp')
 
 cc = meson.get_compiler('c')
 cxx = meson.get_compiler('cpp')


c/meson.build
```diff
@@ -144,15 +144,15 @@ nns_capi_pipeline_dep = declare_dependency(link_with: nns_capi_pipeline_lib)
```
myungjoo, line 144: Your commit obsoletes the need for ```nns_capi_[single/pipeline]_lib```. Can you remove LINE 139-143, 115-119 and update other lines using them?

myungjoo, line 144: Your commit obsoletes the need for ```nns_capi_[single/pipeline]_lib```. Can you remove LINE 139-143, 115-119 and update other lines using them?

meson.build
```diff
@@ -3,13 +3,17 @@ project('ml-api', 'c', 'cpp',
...
+    'b_asneeded=false',
     'werror=true',
     'warning_level=1',
     'c_std=gnu89',
     'cpp_std=c++14'
   ]
 )
 
+add_global_link_arguments('-Wl,--no-as-needed', language: 'c')
+add_global_link_arguments('-Wl,--no-as-needed', language: 'cpp')
```
again4you, line 15: `--as-needed` is set as default in Tizen build infra to reduce the number of symbols in the dynamic section.
However, in this case, it is reasonable for me.

again4you (approved):
After applying this patch, I verified that both single and pipeline are located in libcapi-nnstreamer.so as below. Nice catch!

```bash
$ ldd libcapi-nnstreamer.so.1.8.0
...
libcapi-nnstreamer-single.so.1 => /home/again4you/nnstreamer_work/test/api/build/c/./libcapi-nnstreamer-single.so.1 (0x00007f0350005000)
libcapi-nnstreamer-pipeline.so.1 => /home/again4you/nnstreamer_work/test/api/build/c/./libcapi-nnstreamer-pipeline.so.1 (0x00007f034ffee000)
...
```

c/meson.build
zhoonit: Yes, updated :)

gichan-jang (approved):
Thanks for fixing this bug. LGTM

myungjoo (approved):

meson.build
eli-schwartz, line 15: If you already added `b_asneeded=false` up above I wonder why you *also* need to specify it as a link argument?

Also, either way you probably want `add_project_link_arguments()` rather than the global version.

zhoonit: At least in `gcc version 9.3.0 (Ubuntu 9.3.0-11ubuntu0~18.04.1) ` removing `as_needed` is not enough as gcc implictly add `--as-needed` option.

without these lines,

build.ninja is compiled to

```
build c/libcapi-nnstreamer.so.1.8.0: c_LINKER  | c/libcapi-nnstreamer-pipeline.so.1.8.0.p/libcapi-nnstreamer-pipeline.so.1.8.0.symbols c/libcapi-nnstreamer-single.so.1.8.0.p/libcapi-nnstreamer-single.so.1.8.0.symbols
 LINK_ARGS = --verbose -Wl,--verbose -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libcapi-nnstreamer.so.1 '-Wl,-rpath,$$ORIGIN/' -Wl,-rpath-link,/data/nnstreamer-api/build/c c/libcapi-nnstreamer-single.so.1.8.0 c/libcapi-nnstreamer-pipeline.so.1.8.0 -Wl,--end-group
```

However, when I add manually add `--verbose, -Wl--verbose` args to build.ninja to see what's actually going on, I was able to observe that `--as-needed` is attached by default.

```
COLLECT_GCC_OPTIONS='-o' 'c/libcapi-nnstreamer.so.1.8.0' '-v' '-shared' '-fPIC' '-mtune=generic' '-march=x86-64'
 /usr/lib/gcc/x86_64-linux-gnu/9/collect2 -plugin /usr/lib/gcc/x86_64-linux-gnu/9/liblto_plugin.so -plugin-opt=/usr/lib/gcc/x86_64-linux-gnu/9/lto-wrapper -plugin-opt=-fresolution=/tmp/ccdZpJ9f.res -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lc -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s --build-id --eh-frame-hdr -m elf_x86_64 --hash-style=gnu --as-needed -shared -z relro -o c/libcapi-nnstreamer.so.1.8.0 /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/crti.o /usr/lib/gcc/x86_64-linux-gnu/9/crtbeginS.o -L/usr/lib/gcc/x86_64-linux-gnu/9 -L/usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu -L/usr/lib/gcc/x86_64-linux-gnu/9/../../../../lib -L/lib/x86_64-linux-gnu -L/lib/../lib -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib -L/usr/lib/gcc/x86_64-linux-gnu/9/../../.. --verbose --no-undefined --start-group -soname libcapi-nnstreamer.so.1 -rpath $ORIGIN/ -rpath-link /data/nnstreamer-api/build/c c/libcapi-nnstreamer-single.so.1.8.0 c/libcapi-nnstreamer-pipeline.so.1.8.0 --end-group -lgcc --push-state --as-needed -lgcc_s --pop-state -lc -lgcc --push-state --as-needed -lgcc_s --pop-state /usr/lib/gcc/x86_64-linux-gnu/9/crtendS.o /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/crtn.o
```

Thus, adding `-Wl,--no-asneeded` is also needed at least in my work environment.

zhoonit: I agree `add_project_link_arguments()` is a better option, I'll make a patch for this.

eli-schwartz: Yeah, interestingly enough it seems like some distros (Debian!) actually patch GCC in order to add --as-needed as an internal linker option, while it would seem to make more sense to enable that in packaging LDFLAGS...

PR #140, opened 2022-03-14
[Refactoring] Increase readability and maintainability

again4you:
To increase both readability and maintainability, this PR does refactoring applying the below techniques.

* Move Method
* Decompose Conditional
* Substitute Algorithm

### Self evaluation:
Build test: [X]Passed [ ]Failed [ ]Skipped
Run test: [X]Passed [ ]Failed [ ]Skipped

#### v1
* First draft

#### v2
* Add note tag that the index of ml_tensor_type_e should be equal to that of tensor_type.
* Add error message in case of invalid type conversion

### DIFF
--- a/c/src/ml-api-inference-internal.c
+++ b/c/src/ml-api-inference-internal.c
@@ -18,6 +18,41 @@
 #include "ml-api-internal.h"
 
 /**
+ * @brief Convert the type from ml_tensor_type_e to tensor_type.
+ * @note This code is based on the same order between NNS type and ML type.
+ * The index should be the same in case of adding a new type.
+ */
+static tensor_type
+convert_tensor_type_from (ml_tensor_type_e type)
+{
+  if (type < ML_TENSOR_TYPE_INT32 || type > ML_TENSOR_TYPE_UINT64) {
+    _ml_error_report
+        ("Failed to convert the type. Input ml_tensor_type_e %d is invalid.",
+        type);
+    return _NNS_END;
+  }
+
+  return (tensor_type) type;
+}
+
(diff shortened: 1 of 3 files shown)


gichan-jang (approved):
LGTM

c/src/ml-api-inference-internal.c
```diff
@@ -17,6 +17,30 @@
 #include "ml-api-inference-internal.h"
 #include "ml-api-internal.h"
 
+/**
+ * @brief Convert the type from ml_tensor_type_e to tensor_type.
+ */
+static tensor_type
+convert_tensor_type_from (ml_tensor_type_e type)
+{
+  if (type < ML_TENSOR_TYPE_INT32 || type > ML_TENSOR_TYPE_UINT64)
```
jaeyun-jung, line 26: 1. currently yes (ml-tensor-type and nns-tensor-type is equal)
For future work, plz add note tag here (same enum value between nns and ml type, for the case when new type is added)
2. IMO, line 27 and 39 need error message about invalid type conversion.

again4you: > 1. currently yes (ml-tensor-type and nns-tensor-type is equal)
>        For future work, plz add note tag here (same enum value between nns and ml type, for the case when new type is added)

You're right. This code is based on the same order between NNS type and ML type.
I will add some comments for adding new type in the future.


> 
>     2. IMO, line 27 and 39 need error message about invalid type conversion.

OK. I will add error message for invalid type conversion.

jihochu (approved):

PR #145, opened 2022-04-06
[Package] Change package dependency.

gichan-jang:
Previously, the single and pipeline API were separated.
Change the pipeline package contain the single package.

Signed-off-by: gichan <gichan2.jang@samsung.com>


jaeyun-jung (approved):

myungjoo (approved):
LGTM.

In the next PR, add a call-dep from libnnstreamer.so to libcapi-ml-inference-single.so

PR #153, opened 2022-04-18
[C-API/Service] Workaround for dbspace failure

again4you:
Because of SMACK control, the application process cannot access the
'/opt/usr/dbspace'. To meet the Tizen M1 release schedule, this patch
uses '/tmp' as database space. It is just a workaround and will be fixed
in the right way.

### DIFF
--- a/c/src/ml-api-service-db.cc
+++ b/c/src/ml-api-service-db.cc
@@ -18,7 +18,7 @@
 #include "ml-service-db.hh"
 #include "ml-api-service.h"
 
-#define ML_DATABASE_PATH      SYS_DB_DIR"/.ml-service-leveldb"
+#define ML_DATABASE_PATH      "/tmp/.ml-service-leveldb"
 
 /**
  * @brief Class for implementation of IMLServiceDB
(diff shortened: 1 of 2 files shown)


c/src/ml-api-service-db.cc
```diff
@@ -18,7 +18,7 @@
 #include "ml-service-db.hh"
 #include "ml-api-service.h"
 
-#define ML_DATABASE_PATH      SYS_DB_DIR"/.ml-service-leveldb"
+#define ML_DATABASE_PATH      "/tmp/.ml-service-leveldb"
```
gichan-jang, line 21: How about leaving the `@todo` message?

gichan-jang (approved):
LGTM

myungjoo (approved):
LGTM for a hotfix for the M1RCx release.
After the emergency SR, try something like:
```
#ifdef DBSPACE_PATH
#define ML_DATABASE_PATH #DBSPACE_PATH
#else
#define ML_DATABASE_PATH SYS_DB_DIR"/.ml-service-leveldb"
#endif
```
and add related scripts (-DDBSPACE_PATH="/tmp/...") in .spec file.

PR #154, opened 2022-04-18
[single] Allow tvn model when using VD_AIFW

niley7464:
This patch allows tvn model when using VD_AIFW framework.

Signed-off-by: yelini-jeong <yelini.jeong@samsung.com>


jaeyun-jung (approved):

again4you (approved):
@niley7464 already checked its operation after fixing it in her place.
Good job!

PR #156, opened 2022-04-20
[Feature] Add ml-service feature check function

gichan-jang:
Add machine learning service feature check function.

Signed-off-by: gichan <gichan2.jang@samsung.com>

### DIFF
--- a/c/src/ml-api-common-tizen-feature-check.c
+++ b/c/src/ml-api-common-tizen-feature-check.c
@@ -22,7 +22,13 @@
 /**
  * @brief Tizen ML feature.
  */
-#define ML_INF_FEATURE_PATH "tizen.org/feature/machine_learning.inference"
+static const gchar *ML_FEATURES[] = {
+  [ML_FEATURE] = "tizen.org/feature/machine_learning",
+  [ML_FEATURE_INFERENCE] = "tizen.org/feature/machine_learning.inference",
+  [ML_FEATURE_TRAINING] = "tizen.org/feature/machine_learning.training",
+  [ML_FEATURE_SERVICE] = "tizen.org/feature/machine_learning.service",
+  NULL
+};
 
 /**
  * @brief Internal struct to control tizen feature support (machine_learning.inference).
@@ -71,10 +77,19 @@ _ml_tizen_set_feature_state (int state)
 }
 
 /**
+ * @brief Get machine learning feature path.
+ */
+const char *
+_ml_tizen_get_feature_path (ml_feature_e ml_feature)
+{
+  return ML_FEATURES[ml_feature];
+}
+
+/**
  * @brief Checks whether machine_learning.inference feature is enabled or not.
  */
 int
-_ml_tizen_get_feature_enabled (void)
+_ml_tizen_get_feature_enabled (ml_feature_e ml_feature)
 {
   int ret;
   int feature_enabled;
@@ -90,11 +105,12 @@ _ml_tizen_get_feature_enabled (void)
     return ML_ERROR_NOT_SUPPORTED;
   } else if (NOT_CHECKED_YET == feature_enabled) {
     bool ml_inf_supported = false;
-    ret =
-        system_info_get_platform_bool (ML_INF_FEATURE_PATH, &ml_inf_supported);
+    ret = system_info_get_platform_bool (ML_FEATURES[ml_feature],
+        &ml_inf_supported);
     if (0 == ret) {
       if (false == ml_inf_supported) {
-        _ml_loge ("machine_learning.inference NOT supported");
+        _ml_loge ("%s feature NOT supported! Enable the feature before "
+            "calling ML.Inference APIs.", ML_FEATURES[ml_feature]);
         _ml_tizen_set_feature_state (NOT_SUPPORTED);
         return ML_ERROR_NOT_SUPPORTED;
       }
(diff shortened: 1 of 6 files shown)


c/src/ml-api-internal.h
```diff
@@ -69,24 +69,30 @@ typedef enum
   SUPPORTED = 1
 } feature_state_t;
 
+/**
+ * @brief Enumeration for machine_learning feature.
+ */
+typedef enum {
+  ML_INF_FEATURE = 0,
```
jaeyun-jung, line 76: recommend: name rule and nntrrainer ML_FEATURE_INFERENCE / ML_FEATURE_TRAINING / ML_FEATURE_SERVICE

again4you: I agreed with @jaeyun-jung 's opinion.

c/src/ml-api-common-tizen-feature-check.c
```diff
@@ -22,7 +22,10 @@
 /**
  * @brief Tizen ML feature.
  */
-#define ML_INF_FEATURE_PATH "tizen.org/feature/machine_learning.inference"
+static const gchar *ML_FEATURES[] = {
+  [ML_INF_FEATURE] = "tizen.org/feature/machine_learning.inference",
+  [ML_SERVICE_FEATURE] = "tizen.org/feature/machine_learning.service"
+};
```
jaeyun-jung, line 28: plz add NULL for iteration

c/src/ml-api-common.c
```diff
@@ -26,7 +26,7 @@ ml_tensors_info_create (ml_tensors_info_h * info)
 {
   ml_tensors_info_s *tensors_info;
 
-  check_feature_state ();
+  check_feature_state (ML_INF_FEATURE);
```
jaeyun-jung, line 29: Need discussion: This is common util function. It needs to check all feature - inference / training / service. (Or, machine-learning?)

again4you: Tensor utility functions such as ml_tensors_info_create() could be used with the NNTrainer module as below.
* https://github.com/nnstreamer/nntrainer/blob/main/api/capi/src/nntrainer.cpp#L1043-L1046
```cpp
  status = ml_tensors_info_create(info);
  if (status != ML_ERROR_NONE) {
    return status;
  }
```

NNTrainer API also checks the feature. However, the name of the macro `check_feature_state()` is exactly the same but NNTrainer defines its own macro instead of using a common function.
* https://github.com/nnstreamer/nntrainer/blob/main/api/capi/include/nntrainer_internal.h#L55-L61
```cpp
#define check_feature_state()                         \
  do {                                                \
    int feature_ret = ml_tizen_get_feature_enabled(); \
    if (ML_ERROR_NONE != feature_ret)                 \
      return feature_ret;                             \
  } while (0);
```

In my opinion, it is not a good idea since a single process has two different logic for checking the Tizen feature. I hope that `check_feature_state()` in ml-api-common.c will be provided as common utility function.

again4you (changes requested):
Unlike in the previous code, we need to check various features. Your work is intuitive and easy to understand so increase readability. 

In my opinion, it would be better if some code is updated to comply with the Naming rule and coding convention.

c/src/ml-api-common-tizen-feature-check.c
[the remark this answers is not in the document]
jaeyun-jung: agreed

gichan-jang:
@jaeyun-jung @again4you Thanks for the reviews. I updated all you mentioned. Please check it again.

jaeyun-jung (approved):

again4you (approved):
All items reviewers commented are updated. 
Good job! 👍

c/src/ml-api-internal.h
```diff
@@ -69,24 +69,30 @@ typedef enum
...
   do { \
-    int feature_ret = _ml_tizen_get_feature_enabled (); \
+    int feature_ret = _ml_tizen_get_feature_enabled (ml_feature); \
     if (ML_ERROR_NONE != feature_ret) { \
-      if (ML_ERROR_NOT_SUPPORTED == feature_ret) \
-        _ml_error_report_return (feature_ret, \
-            "Tizen MachineLearning.Inference feature is not enabled. Enable the feature before calling ML.Inference APIs."); \
       _ml_error_report_return (feature_ret, \
           "_ml_tizen_get_feature_enabled() has returned an error. (%d)", \
           feature_ret); \
```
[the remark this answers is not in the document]
gichan-jang: Right. It's not good error message for the general users. I modified the error message a little more elaborately as you guided.

c/src/ml-api-common-tizen-feature-check.c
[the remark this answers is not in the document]
gichan-jang: Nice suggestion. I added NNTrainer feature. Thanks.

c/src/ml-api-common.c
```diff
@@ -51,7 +51,7 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
...
```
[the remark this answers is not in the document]
gichan-jang: Right! With this change, feature checks can be generally used from other users who want to use ml-api.

c/src/ml-api-common.c
gichan-jang: Right. It can be commonly used in the NNTrainer.
Using a common function is more helpful for code management and debugging.
I will update the NNTrainer for feature check with these changes. Thanks.

PR #157, opened 2022-04-21
[TDD] Add Mock Testcase for dependency module & do refactoring for Service API

again4you:
To increase the test coverage, this PR adds the Mock test cases for the external database module. During this process, I found out that some embedded database does not support multi-process and it might cause the IO exception when multiple clients write the key simultaneously. This PR also does refactoring for handling this issue by closing the DB handle explicitly after use and it can reduce the deadlock situation.

### UML
uml

### Future work
If the `daemon` process at the system level only accesses the database, it would be fine.


tests/capi/unittest_capi_service_db_mock.cc
```diff
@@ -0,0 +1,477 @@
...
+
+/**
+ * @brief Check the normal process of setting the pipeline description with a given name.
+ */
+TEST (MLServiceAPIDBMockTest, set_pipeline_description_0_p)
+{
+  int ret = 0;
+  const gchar *key = "ServiceName";
+  gchar *pipeline = g_strdup_printf ("appsrc name=appsrc ! "
+      "other/tensor,dimension=(string)1:1:1:1,type=(string)float32,framerate=(fraction)0/1 ! "
```
gichan-jang, line 157: Since `ohter/tensors` is deprecated, it would be better using `other/tensors`.

gichan-jang, line 157: Since `ohter/tensors` is deprecated, it would be better using `other/tensors`.

tests/capi/unittest_capi_service_db_mock.cc
```diff
@@ -0,0 +1,477 @@
...
+ * @brief Check the error handling when leveldb_open() raises an IO exception.
+ */
+TEST (MLServiceAPIDBMockTest, leveldb_open_IO_ERROR_n)
+{
+  int ret = 0;
+  const gchar *key = "ServiceName";
+  gchar *pipeline = g_strdup_printf ("appsrc name=appsrc ! "
+      "other/tensor,dimension=(string)1:1:1:1,type=(string)float32,framerate=(fraction)0/1 ! "
+      "tensor_sink name=tensor_sink");
+  mockInstance = new LevelDBMock();
```
gichan-jang, line 186: These lines are repeated. How about using test fixture?

gichan-jang, line 186: These lines are repeated. How about using test fixture?

tests/capi/unittest_capi_service_db_mock.cc
```diff
@@ -0,0 +1,477 @@
...
+  EXPECT_CALL (*mockInstance, leveldb_close(_))
+      .WillRepeatedly(Return());
+
+  /* Test */
+  ret = ml_service_set_pipeline (key, pipeline);
+  EXPECT_EQ (ML_ERROR_NONE, ret);
+
+  g_free (pipeline);
+  delete mockInstance;
+}
```
gichan-jang, line 174: Don't you need to delete the pipeline description before finishing the test?
If not, next test is affected by previous test. I think the test should be independent from other tests.

gichan-jang, line 174: Don't you need to delete the pipeline description before finishing the test?
If not, next test is affected by previous test. I think the test should be independent from other tests.

tests/capi/unittest_capi_service_db_mock.cc
```diff
@@ -0,0 +1,477 @@
...
+      "other/tensor,dimension=(string)1:1:1:1,type=(string)float32,framerate=(fraction)0/1 ! "
+      "tensor_sink name=tensor_sink");
+  mockInstance = new LevelDBMock();
+
+  EXPECT_CALL (*mockInstance, leveldb_open(_, _, _))
+      .Times(1).WillOnce(Return((leveldb_t*)0x1234));
+  EXPECT_CALL (*mockInstance, leveldb_put(_, _, _, _, _, _, _))
+      .Times(1).WillOnce(SetArgPointee<6>(nullptr));
+  EXPECT_CALL (*mockInstance, leveldb_close(_))
+      .WillRepeatedly(Return());
```
gichan-jang, line 166: Can't we know how many times this function will be called?

gichan-jang, line 166: Can't we know how many times this function will be called?

tests/capi/unittest_capi_service_db_mock.cc
again4you: Nice catch. I updated the tensor type.

again4you: Updated. Thanks.

tests/capi/unittest_capi_service_db_mock.cc
again4you: Right. I applied Test Fixture as below. We can remove redundant code in each test case.

```cpp
class MLServiceAPIDBMockTest : public ::testing::Test
{
protected:
  const gchar *key;
  const gchar *invalid_key;
  gchar *pipeline;
  gsize len;

  void SetUp () override
  {
    key = "ServiceName";
    invalid_key = "InvalidKey";
    pipeline = g_strdup_printf ("appsrc name=appsrc ! "
      "other/tensors,dimension=(string)1:1:1:1,type=(string)float32,framerate=(fraction)0/1 ! "
      "tensor_sink name=tensor_sink");
    len = strlen (pipeline);
  }
  void TearDown () override
  {
    g_free (pipeline);
  }
};
```

again4you: Fixed. Thanks.

tests/capi/unittest_capi_service_db_mock.cc
again4you: I added the logic for verifying the count of `leveldb_close()`. Thanks.

again4you: Updated!

tests/capi/unittest_capi_service_db_mock.cc
again4you: Right. I applied the Test Fixture so each unit test is independent of others.

again4you: Fixed.

c/src/ml-api-service-db.cc
```diff
@@ -43,7 +44,7 @@ class MLServiceLevelDB:public IMLServiceDB
 private:
   MLServiceLevelDB (std::string path);
   virtual ~MLServiceLevelDB ();
-
+
```
myungjoo, line 47: style error?

packaging/machine-learning-api.spec
```diff
@@ -321,6 +322,9 @@ bash %{test_script} ./tests/capi/unittest_capi_inference
 bash %{test_script} ./tests/capi/unittest_datatype_consistency
 bash %{test_script} ./tests/capi/unittest_capi_service
 
+# bash %{test_script} ./tests/capi/unittest_capi_service
```
myungjoo, line 325: Meaningless line?

c/src/ml-api-service-db.cc
again4you: Fixed. Thank you for notifying me.

packaging/machine-learning-api.spec
again4you: Oops! My mistake. Fixed.

gichan-jang (approved):
LGTM!

myungjoo (approved):
LGTM

PR #158, opened 2022-04-22
[C-Api] handle each feature state

jaeyun-jung:
Each Tizen ML feature state (inference, training, service) should be managed separately.

### DIFF
--- a/c/src/ml-api-common-tizen-feature-check.c
+++ b/c/src/ml-api-common-tizen-feature-check.c
@@ -27,7 +27,7 @@ static const gchar *ML_FEATURES[] = {
   [ML_FEATURE_INFERENCE] = "tizen.org/feature/machine_learning.inference",
   [ML_FEATURE_TRAINING] = "tizen.org/feature/machine_learning.training",
   [ML_FEATURE_SERVICE] = "tizen.org/feature/machine_learning.service",
-  NULL
+  [ML_FEATURE_MAX] = NULL
 };
 
 /**
@@ -37,7 +37,7 @@ static const gchar *ML_FEATURES[] = {
 typedef struct
 {
   GMutex mutex;
-  feature_state_t feature_state;
+  feature_state_t feature_state[ML_FEATURE_MAX];
 } feature_info_s;
 
 static feature_info_s *feature_info = NULL;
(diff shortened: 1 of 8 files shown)


c/src/ml-api-common-tizen-feature-check.c
```diff
@@ -97,25 +100,27 @@ _ml_tizen_get_feature_enabled (ml_feature_e ml_feature)
...
   g_mutex_lock (&feature_info->mutex);
-  feature_enabled = feature_info->feature_state;
+  feature_enabled = feature_info->feature_state[ml_feature];
   g_mutex_unlock (&feature_info->mutex);
 
   if (NOT_SUPPORTED == feature_enabled) {
-    _ml_loge ("machine_learning.inference NOT supported");
+    _ml_loge ("Tizen machine learning feature (%s) is NOT supported.",
+        ML_FEATURES[ml_feature]);
     return ML_ERROR_NOT_SUPPORTED;
```
gichan-jang, line 109: Minor: How about using `_ml_error_report_return`?

gichan-jang (approved):
LGTM

c/src/ml-api-common-tizen-feature-check.c
jaeyun-jung: Internally we use the macro check_feature_state() and this will set error message in ml-error-report.

again4you (approved):
Good!

c/src/ml-api-internal.h
```diff
@@ -73,12 +73,12 @@ typedef enum
...
  */
 typedef enum {
-  ML_FEATURE  = 0,
+  ML_FEATURE = 0,
   ML_FEATURE_INFERENCE,
   ML_FEATURE_TRAINING,
   ML_FEATURE_SERVICE,
 
-  ML_FEATURE_UNKNOWN
+  ML_FEATURE_MAX
```
myungjoo, line 81: If you have additional FEATURE enum, MAX will be updated. Be sure that ML_FEATURE_MAX won't be exposed to public API and public apps won't have this value embedded in their binaries.

c/src/ml-api-common-tizen-feature-check.c
```diff
@@ -97,46 +100,50 @@ _ml_tizen_get_feature_enabled (ml_feature_e ml_feature)
   ml_tizen_initialize_feature_state ();
 
   g_mutex_lock (&feature_info->mutex);
-  feature_enabled = feature_info->feature_state;
+  feature_enabled = feature_info->feature_state[ml_feature];
   g_mutex_unlock (&feature_info->mutex);
 
   if (NOT_SUPPORTED == feature_enabled) {
-    _ml_loge ("machine_learning.inference NOT supported");
+    _ml_loge ("Tizen machine learning feature (%s) is NOT supported.",
```
myungjoo, line 107: Please add what the developer should do with this error.

E.g.,
Tizen machine learning feature (%s) is NOT supported. Please check if your application has properly requested the feature and the device has the feature installed.

myungjoo (approved):
LGTM.

@jaeyun-jung please take a look at the comments before you merge this.

myungjoo (approved):
Good2Go!

PR #191, opened 2022-08-18
[ml-agent] Add set/get/delete pipeline method to pipeline module

anyj0527:
- Change the method name `register_pipeline` -> `launch_pipeline`.
  This method launch pipeline with given service name, if the pipeline
  description is stored in the DB beforehand.
- Change the module source file into cpp file. It's to use the db implementation.
- Add `set_pipeline`, `get_pipeline`, and `delete_pipeline` dbus call.
- These correspond with `ml_service_[set|get|delete]_pipeline` APIs.


again4you (approved):
LGTM!

myungjoo (approved):

PR #194, opened 2022-08-22
[ml-service] Fix session type of ml-service API @open sesame 08/22 18:35

anyj0527:
- Set System type for ml-service API.
- If fails to get system bus, try with session type (for unittest)

**Self evaluation:**

### DIFF
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -145,6 +145,9 @@ ml_service_delete_pipeline (const char *name)
   return ret;
 }
 
+/**
+ * @brief Launch the pipeline of given service.
+ */
 int
 ml_service_launch_pipeline (const char *name, ml_service_h * h)
 {
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -31,12 +31,24 @@ typedef struct
 static MachinelearningServicePipeline *
 _get_proxy_new_for_bus_sync (void)
 {
+  MachinelearningServicePipeline *mlsp;
+
   /** @todo deal with GError */
-  return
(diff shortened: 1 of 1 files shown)


c/src/ml-api-service-agent-client.c
```diff
@@ -31,12 +31,24 @@ typedef struct
 static MachinelearningServicePipeline *
 _get_proxy_new_for_bus_sync (void)
 {
+  MachinelearningServicePipeline *mlsp;
+
   /** @todo deal with GError */
-  return
-      machinelearning_service_pipeline_proxy_new_for_bus_sync
+  mlsp = machinelearning_service_pipeline_proxy_new_for_bus_sync
+      (G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE,
```
jaeyun-jung, line 38: Q: is it possible to check bus-type? (availability of G_BUS_TYPE_SYSTEM or G_BUS_TYPE_SESSION)

anyj0527: Well, I think checking the return value of this function seems proper to do that.

again4you: In my opinion, this patch is OK as `Workaround`.
The reason why try both `G_BUS_TYPE_SYSTEM` and `G_BUS_TYPE_SESSION` is that our CI test uses `Session Bus` but Tizen devices use `System Bus`. In this case, we build two libraries: One is for Device, the other is for CI Test.

Anyway, to submit the platform image, it's ok.

gichan-jang (approved):
LGTM

again4you (approved):
As a workaround, LGTM!

jaeyun-jung (approved):

PR #200, opened 2022-09-01
[ml-agent] Add prefix for key in daemon's DB @open sesame 9/1 18:18

anyj0527:
- To prevent storing plain keys in DB, hash it.
- Add prefix for key when access DB
- This prefix could be set in build time via meson option.


again4you (approved):
LGTM!

jaeyun-jung (approved):

PR #202, opened 2022-09-05
[svace][trivial] Fix svace issues for leak and uninitialize

anyj0527:
- Free allocated `dllist` when allocation of other list is failed.
- Initialize an uninitialized value.


jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #203, opened 2022-09-05
[service] Add a guide for the "key" when using ml_service_set_pipeline

anyj0527:
- Add a sentence which warns users to keep their name (key) safe.


myungjoo (approved):

anyj0527:
Note: Adding document do not need an ACR process.

again4you (approved):
LGTM!

PR #205, opened 2022-09-06
[Build] dependency to dlog

jaeyun-jung:
Add dlog dependency when building daemon on Tizen release.


again4you (approved):
LGTM!

PR #216, opened 2022-09-28
[agent] Fix occasional hanging issue when destroy certain pipelines

anyj0527:
- When pipeline is destroyed, agent tried to set its state as GST_STATE_NULL.
- However, pipelines which contain tensor_query_* elements make hanging
  when change te state. To resolve this, as a workaround, this commit just unref it.
- After fix this issue in tensor_query / nnstreamer-edge, revive this code.
- Remove lock/unlock when freeing pipeline instances


jaeyun-jung (approved):

again4you (approved):
LGTM!

myungjoo (approved):
LGTM.

For that free mechanism, "lock" should've been a pointer in the struct.

PR #218, opened 2022-10-04
[C-Api] feature for float16 type

jaeyun-jung:
Code clean, update feature name (same with nnstreamer repo) and size of float16 type.


c/src/ml-api-common.c
```diff
@@ -387,12 +387,12 @@ ml_tensors_info_set_tensor_type (ml_tensors_info_h info,
         "The parameter, type, ML_TENSOR_TYPE_UNKNOWN or out of bound. The value of type should be between 0 and ML_TENSOR_TYPE_UNKNOWN - 1. type = %d, ML_TENSOR_TYPE_UNKNOWN = %d.",
         type, ML_TENSOR_TYPE_UNKNOWN);
 
-#ifndef SUPPORT_FLOAT16
+#ifndef FLOAT16_SUPPORT
```
myungjoo, line 390: Do we have this macro in our build script of API? (FLOAT16_SUPPORT is correct, though)

jaeyun-jung: No, we don't have this yet.

myungjoo (approved):

PR #231, opened 2023-01-02
[Test] fix double free case

jaeyun-jung:
Fix double free case in ml-service testcases.


again4you (approved):
GLTM!

In my humble opinion, we need defensive programming such as setting the gchar pointer as NULL before use.
```cpp
ret_pipeline = nullptr;
status = ml_service_get_pipeline (service_name, &ret_pipeline);
```

gichan-jang (approved):
LGTM

PR #239, opened 2023-01-17
[Single] prevent use-after-free case

jaeyun-jung:
Clone input data in single-shot, to prevent use-after-free case when invoking a model with timed out option.

Signed-off-by: Jaeyun Jung <jy1210.jung@samsung.net>

### DIFF
--- a/c/src/ml-api-common.c
+++ b/c/src/ml-api-common.c
@@ -763,6 +763,47 @@ _ml_tensors_data_clone_no_alloc (const ml_tensors_data_s * data_src,
 }
 
 /**
+ * @brief Copies the tensor data frame.
+ */
+int
+ml_tensors_data_clone (const ml_tensors_data_h in, ml_tensors_data_h * out)
+{
+  int status;
+  unsigned int i;
+  ml_tensors_data_s *_in, *_out;
+
+  check_feature_state (ML_FEATURE);
+
+  if (in == NULL)
+    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
+        "The parameter, in, is NULL. It should be a valid ml_tensors_data_h handle, which is usually created by ml_tensors_data_create ().");
+
+  if (out == NULL)
+    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
+        "The parameter, out, is NULL. It should be a valid pointer to ml_tensors_data_h handle.");
+
+  _in = (ml_tensors_data_s *) in;
+  G_LOCK_UNLESS_NOLOCK (*_in);
+
+  status = ml_tensors_data_create (_in->info, out);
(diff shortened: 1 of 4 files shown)


c/src/ml-api-common.c
```diff
@@ -762,6 +762,47 @@ _ml_tensors_data_clone_no_alloc (const ml_tensors_data_s * data_src,
...
+  _in = (ml_tensors_data_s *) in;
+  G_LOCK_UNLESS_NOLOCK (*_in);
+
+  status = ml_tensors_data_create (_in->info, out);
+  if (status != ML_ERROR_NONE) {
+    _ml_loge ("Failed to create new handle to copy tensor data.");
+    goto error;
+  }
+
+  _out = (ml_tensors_data_s *) (*out);
```
myungjoo, line 794: POINTER ERROR

myungjoo (changes requested):
There's a pointer error.

myungjoo (approved):
LGTM

c/src/ml-api-inference-single.c
```diff
@@ -1385,7 +1389,14 @@ _ml_single_invoke_internal (ml_single_h single,
...
   }
 
-  single_h->input = input;
+  /**
+   * Clone input data here to prevent use-after-free case.
+   * We should release single_h->input after calling __invoke() function.
+   */
+  status = ml_tensors_data_clone (input, &single_h->input);
+  if (status != ML_ERROR_NONE)
+    goto exit;
```
myungjoo, line 1398: Is this memcpy really required?

The invoke APIs are synchronous.

jaeyun-jung: The exception case when releasing input data handle while invoking a model. Now tf-lite subplugin directly uses the input data in single-shot.

c/src/ml-api-common.c
```diff
@@ -762,6 +762,47 @@ _ml_tensors_data_clone_no_alloc (const ml_tensors_data_s * data_src,
...
+
+  for (i = 0; i < _out->num_tensors; ++i) {
+    memcpy (_out->tensors[i].tensor, _in->tensors[i].tensor,
```
songgot, line 797: I wonder if there is a case where '_in' is not nullptr but '_in->tensors[i].tensor' is nullptr.
Thank you.

jaeyun-jung: User should create data handle using ml_tensors_data_create(), and it checks allocation failure case. (plz see line 845)

songgot: Thanks for the detailed explanation.
Thank you.

songgot (approved):
LGTM

again4you (approved):
LGTM!

PR #240, opened 2023-01-26
[C-Api/Single] invalid data handle in invoke thread @open sesame 01/26 17:26

jaeyun-jung:
Prevent error case - null data ptr in invoke thread.

Signed-off-by: Jaeyun Jung <jy1210.jung@samsung.com>


songgot (approved):
LGTM

anyj0527 (approved):
LGTM✔

PR #248, opened 2023-02-10
[dbus] Add DBus interface for model management

anyj0527:
- Add Register, Update, Rollback, SetInference, Get, GetLatest interfaces. These will be used by new ml-service APIS for model management.
- Remove outdated dbus call for model (get_path, set_path, delete) and tests.


dbus/model-dbus.xml
```diff
@@ -18,5 +18,33 @@
       <arg type="s" name="name" direction="in" />
       <arg type="i" name="result" direction="out" />
     </method>
+    <!-- Register the given model -->
+    <method name="Register">
```
jaeyun-jung, line 22: Discussion: We didnt get the ACR process for ml-service model yet.
If the interfaces - set-path/get-path/delete are unnecessary, let's remove these.

anyj0527: You're right. I removed the outdated methods and related tests. PTAL

again4you (approved):
LGTM!

jaeyun-jung (approved):

PR #258, opened 2023-02-22
[agent][DB] Fix "delete_pipeline" to check the number of changed rows

anyj0527:
- Check the number of changed rows after the DELETE query
- Add some negative testcases.


again4you (approved):
Good job!

PR #262, opened 2023-03-09
[Service] handle table version

jaeyun-jung:
Handle table version. If table schema is changed, we should implement data migration later.


songgot (approved):
I don't fully understand what g_strdup_printf prints TBL_VER_PIPELINE_DESCRIPTION and TBL_VER_MODEL_INFO respectively. 
but LGTM.

PR #265, opened 2023-03-13
[Service] validate model path

jaeyun-jung:
Validate model path when registering a model to database. It should be an absolute path and not symbol link file.


anyj0527 (approved):
LGTM✔I'll add negative TC for the permission error.

songgot (approved):
LGTM

anyj0527:
I think it's good to go

PR #269, opened 2023-03-22
[agent][trivial] Print exception and fix memleak

anyj0527:
- Print exception message rather than simple error log
- Add missing `ml_option_destroy` in a unittest


songgot (approved):
LGTM

jaeyun-jung (approved):

PR #271, opened 2023-03-24
[service] Change error code not_support into io_error

anyj0527:
- If failed to connect with dbus, return ML_ERROR_IO_ERROR
  rather than ML_ERROR_NOT_SUPPORTED, which should be check with tizen
  feature.
- Add negative TC with DBUS unconnected.


jaeyun-jung (approved):
LGTM. I will make new PR about _destroy() function, plz review it later.

PR #273, opened 2023-03-24
[Service] do not free handle in error case

jaeyun-jung:
If called service-destroy function, do not free allocated handle if an error from dbus or database is raised.


songgot (approved):
Looks good to me.

anyj0527 (approved):
LGTM✔

PR #274, opened 2023-03-27
[Build/Android] temporal block flatbuffer

jaeyun-jung:
Fix error, temporal block flatbuffer when building android library.


niley7464 (approved):
LGTM :)

PR #275, opened 2023-03-27
[Build/Test] daemon for unittest

jaeyun-jung:
Cleanup build script, move test-daemon into 'tests' dir and fix file names.


daemon/meson.build
```diff
@@ -1,105 +1,80 @@
...
-  if get_option('enable-test')
-    gdbus_gen_test_src = custom_target('gdbus-gencode-test',
-      input : test_dbus_input,
-      output : ['test-dbus.h', 'test-dbus.c'],
-      command : [gdbus_prog, '--interface-prefix', 'org.tizen',
-                '--generate-c-code', 'test-dbus',
-                '--output-directory', meson.current_build_dir(),
-                '@INPUT@'])
+# Generate GDbus header and code
+gdbus_prog = find_program('gdbus-codegen', required : true)
```
songgot, line 11: If require is true meson will abort if no program can be found.
I understand that this was intended.

songgot (approved):
Looks good to me.

daemon/meson.build
jaeyun-jung: Right, ml-service daemon requires this.

PR #278, opened 2023-03-29
[CodeClean] Remove unnecessary code in ml-agent

anyj0527:
- Move pipeline related code in gdbus-util to proper src.
- Let std::exception catch those runtime errors.
- Remove unnecessary code in service api test

### DIFF
--- a/daemon/gdbus-util.c
+++ b/daemon/gdbus-util.c
@@ -115,34 +115,6 @@ gdbus_disconnect_signal (gpointer instance, int num_signals,
 }
 
 /**
- * @brief Cleanup the instance of the DBus interface.
- */
-static void
-put_instance (gpointer * instance)
-{
-  g_object_unref (*instance);
-  *instance = NULL;
-}
-
-/**
- * @brief Get the skeleton object of the DBus interface.
- */
-MachinelearningServicePipeline *
-gdbus_get_instance_pipeline (void)
-{
-  return machinelearning_service_pipeline_skeleton_new ();
-}
-
-/**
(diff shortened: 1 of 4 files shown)


daemon/gdbus-util.c
```diff
@@ -117,31 +117,13 @@ gdbus_disconnect_signal (gpointer instance, int num_signals,
 /**
  * @brief Cleanup the instance of the DBus interface.
  */
-static void
+void
 put_instance (gpointer * instance)
```
jaeyun-jung, line 121: we can remove this function also :)

daemon/pipeline-dbus-impl.cc
```diff
@@ -68,6 +69,24 @@ static void _pipeline_free (gpointer data)
   g_free (p);
 }
 
+/**
+ * @brief Get the skeleton object of the DBus interface.
+ */
+MachinelearningServicePipeline *
+gdbus_get_instance_pipeline (void)
```
jaeyun-jung, line 76: line 76 and 85, it would be a static.
Also, plz consider changing function name (gdbus_get_pipeline_instance / gdbus_put_pipeline_instance) same rule in model-dbus)

daemon/gdbus-util.c
anyj0527: You're right. I removed this function.

daemon/pipeline-dbus-impl.cc
anyj0527: I renamed those functions, PTAL

jaeyun-jung (approved):

PR #241, opened 2023-01-27
[C-API] Increase ML_TENSOR_RANK_LIMIT @open sesame 04/06 11:25

niley7464:
This patch increases ML_TENSOR_RANK_LIMIT from 4 to 16.

Related to : https://github.com/nnstreamer/nnstreamer/pull/3987

Signed-off-by: Yelin Jeong <yelini.jeong@samsung.com>

PR #283, opened 2023-04-03
[Single] Add input and output ranks @open sesame 4/3 15:31

niley7464:
This patch adds input and output ranks to ml_single. 
Input and output ranks are used when user explicitly sets rank property.
Related to https://github.com/nnstreamer/api/pull/241#issuecomment-1489867402

### DIFF
--- a/c/src/ml-api-inference-single.c
+++ b/c/src/ml-api-inference-single.c
@@ -134,6 +134,10 @@ typedef struct
   ml_tensors_data_s in_tensors;    /**< input tensor wrapper for processing */
   ml_tensors_data_s out_tensors;   /**< output tensor wrapper for processing */
 
+  /** @todo Use only ml_tensor_info_s dimension instead of saving ranks value */
+  guint input_ranks[ML_TENSOR_SIZE_LIMIT];   /**< the rank list of input tensors, it is calculated based on the dimension string. */
+  guint output_ranks[ML_TENSOR_SIZE_LIMIT];  /**< the rank list of output tensors, it is calculated based on the dimension string. */
+
   GList *destroy_data_list;         /**< data to be freed by filter */
 } ml_single;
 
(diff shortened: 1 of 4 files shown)


c/src/ml-api-inference-single.c
```diff
@@ -134,6 +134,10 @@ typedef struct
   ml_tensors_data_s in_tensors;    /**< input tensor wrapper for processing */
   ml_tensors_data_s out_tensors;   /**< output tensor wrapper for processing */
 
+  /** @todo Use only ml_tensor_info_s dimension instead of saving ranks value */
+  unsigned int input_ranks[ML_TENSOR_SIZE_LIMIT];   /**< the rank list of input tensors, it is calculated based on the dimension string. */
```
wooksong, line 138: How about using ```guint``` for code consistency?

niley7464: updated it!
Thank you :)

c/src/ml-api-inference-single.c
```diff
@@ -1745,8 +1749,33 @@ ml_single_set_property (ml_single_h single, const char *name, const char *value)
       num = gst_tensors_info_parse_types_string (&gst_info, value);
     else if (g_str_has_suffix (name, "name"))
       num = gst_tensors_info_parse_names_string (&gst_info, value);
-    else
-      num = gst_tensors_info_parse_dimensions_string (&gst_info, value);
+    else {
```
songgot, line 1752: It looks good.
It seems to be one purpose to bring a dimension to the rank[] list according to input or output.
It would be nice to make it a function as well.
static
gint
xxx(GstTensorsInfo * info, const gchar * dim_string, guint *rank)
{

 return num_dim;
}

c/src/ml-api-inference-single.c
```diff
@@ -1821,6 +1849,40 @@ ml_single_get_property (ml_single_h single, const char *name, char **value)
     /* boolean */
     g_object_get (G_OBJECT (single_h->filter), name, &bool_value, NULL);
     *value = (bool_value) ? g_strdup ("true") : g_strdup ("false");
+  } else if (g_str_equal (name, "input") || g_str_equal (name, "output")) {
+    gchar *dim_str = NULL;
```
songgot, line 1853: How about creating a function?

jaeyun-jung: Suggestion: TBU, adding 'rank' in GstTensorInfo, so that we can provide exact dim-string from nnstreamer side, not ML-API inside.

jaeyun-jung (approved):

c/src/ml-api-inference-single.c
songgot: okay, thank you.

songgot (approved):
LGTM

PR #288, opened 2023-04-13
[C-Api] Revert fill dim (default 1) commit

again4you:
Because of native TC failure, this patch reverts the fill dim (default 1) commit.


niley7464 (approved):
LGTM!
Thank you

jaeyun-jung (approved):

PR #292, opened 2023-04-17
[pkgmgr] Support "rpk" install

anyj0527:
- When rpk is installed, parse the provided "model_description.json" file
  and register models described in the json.

TODO: Handle when the rpk package is removed or updated.


daemon/includes/pkg-mgr.h
```diff
@@ -18,17 +18,16 @@
...
-#endif /* __cplusplus */
-
 #if defined(__TIZEN__)
 #include <glib.h>
 #include <stdio.h>
 
 #include "log.h"
 #include "package_manager.h"
 
+#ifdef __cplusplus
```
jaeyun-jung, line 28: weird change..

anyj0527: Updated, PTAL.

again4you (approved):
Basically, LGTM!

Trivial Question)
Should we consider `PACKAGE_MANAGER_EVENT_TYPE_RES_COPY` event?

jaeyun-jung:
plz rebase this also

anyj0527:
> Basically, LGTM!
> 
> Trivial Question) Should we consider `PACKAGE_MANAGER_EVENT_TYPE_RES_COPY` event?

Well I think we need to when these events could be used, the doc is saying that these are related with resource package. We may ask to the person in charge of package_manager.

anyj0527:
> plz rebase this also

Rebased, Thanks!

jaeyun-jung (approved):

PR #293, opened 2023-04-17
[service/model] Check and parse app_info of tizen application (the caller)

anyj0527:
In a Tizen application context,
- `ml_service_model_register` fill app_info column. RPK cannot call this API, so the value of key `is_rpk` is always false.
- `ml_service_model_get*` APIs should parse the app_info and update the path in case of models from 'RPK'.

### DIFF
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -17,6 +17,82 @@
 #include "ml-api-service.h"
 #include "ml-api-service-private.h"
 
+#if defined(__TIZEN__)
+#include <app_common.h>
+
+/**
+ * @brief Parse app_info and update path (for model from rpk). Only for Tizen Applications.
+ */
+static int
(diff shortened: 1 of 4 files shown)


c/src/ml-api-service-agent-client.c
```diff
@@ -17,6 +17,84 @@
 #include "ml-api-service.h"
 #include "ml-api-service-private.h"
 
+#if defined(__TIZEN__)
+#include <app_common.h>
+#endif
+
+#if defined(__TIZEN__)
```
jaeyun-jung, line 24: you can remove line 22~24, same feature.

anyj0527: Thanks, I removed those lines. PTAL

jaeyun-jung:
need rebase

anyj0527:
> need rebase

Rebased, PTAL.

again4you (approved):
LGTM!

jaeyun-jung (approved):

PR #314, opened 2023-05-02
[Service] handle invalid version @open sesame 05/03 09:30

jaeyun-jung:
Raise exception when given version number is 0.


gichan-jang (approved):
LGTM

PR #319, opened 2023-05-03
[Service/Log] printing internal value

jaeyun-jung:
The ID from pipeline hashtable in daemon is internal value. This PR hides this internal ID value.


anyj0527 (approved):
LGTM 👍

PR #332, opened 2023-05-22
API/Client: Revise null-checks for output parameters 

wooksong:
This patch eliminates null-checks for the output parameters, which require memory allocation at the callee side.

### DIFF
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -18,6 +18,9 @@
 #include "ml-api-service-private.h"
 #include "ml-api-service.h"
 
+#define WARN_MSG_DPTR_SET_OVER "The memory blocks pointed by pipeline_desc will be set over with a new one.\n" \
+        "It is highly suggested that `%s` before it is set."
+
 #if defined(__TIZEN__)
 #include <app_common.h>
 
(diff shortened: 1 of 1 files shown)


c/src/ml-api-service-agent-client.c
```diff
@@ -143,9 +143,8 @@ ml_service_get_pipeline (const char *name, char **pipeline_desc)
         "The parameter, 'name' is NULL, It should be a valid string");
   }
 
-  if (!pipeline_desc) {
-    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
-        "The parameter 'pipeline_desc'. It should be a valid char**");
+  if (pipeline_desc) {
+    *pipeline_desc = NULL;
```
songgot, line 147: I don't know well.
Shouldn't 'pipeline_desc' be the first address where memory is allocated?

wooksong: This is a **[out]** parameter.

An example usage as follows:
```
char *desc;

ml_service_get_pipeline(name, &desc);
```

```desc``` will point the first block of an allocated memory inside wherever ```ml_service_get_pipeline```.

**Review by songgot — APPROVED** [review]
LGTM

File: c/src/ml-api-service-agent-client.c (hunk 1 above)
Comment (songgot) [reply c/src/ml-api-service-agent-client.c:147]: Thank you for the detailed description.

File: c/src/ml-api-service-agent-client.c [hunk 2]
```diff
@@ -318,6 +339,7 @@ ml_service_get_pipeline_state (ml_service_h h, ml_pipeline_state_e * state)
   g_clear_error (&err);
 
   *state = (ml_pipeline_state_e) _state;
+
```
myungjoo, line 342: nah..

myungjoo (approved):
LGTM, except for the empty line add.

c/src/ml-api-service-agent-client.c
wooksong: Removed. Thanks.

niley7464 (approved):
LGTM 👍

jaeyun-jung (approved):

PR #333, opened 2023-05-22
API/Client: Replace g_new0 with g_try_new0

wooksong:
g_new0 would not work as the author intended. This patch replaces g_new0 macros with g_try_new0 macros so that memory allocation failure is checked at runtime.


songgot (approved):
LGTM

myungjoo (approved):
LGTM.

niley7464 (approved):
LGTM :)

jaeyun-jung (approved):

PR #334, opened 2023-05-22
API/Client: Use nested scope for JSON handling 

wooksong:
This is a trivial patch for code cleaning. Toward better readability, nested scope and g_auto* are used for JSON handling.

### DIFF
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -497,7 +497,6 @@ int
 ml_service_model_activate (const char *name, const unsigned int version)
 {
   int ret = ML_ERROR_NONE;
-
   GError *err = NULL;
 
   check_feature_state (ML_FEATURE_SERVICE);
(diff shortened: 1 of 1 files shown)


c/src/ml-api-service-agent-client.c
```diff
@@ -143,9 +143,8 @@ ml_service_get_pipeline (const char *name, char **pipeline_desc)
         "The parameter, 'name' is NULL, It should be a valid string");
   }
 
-  if (!pipeline_desc) {
-    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
-        "The parameter 'pipeline_desc'. It should be a valid char**");
+  if (pipeline_desc) {
```
myungjoo, line 146: ml_agent_dbus_interface_pipeline_get_description does the null check of pipeline_desc?

wooksong: @myungjoo

```ml_agent_dbus_interface_pipeline_get_description``` does, and it would be a problem only in the case of ```ml_agent_dbus_interface_pipeline_get_description(..., NULL, ...)```.

songgot (approved):
LGTM

myungjoo (approved):

anyj0527 (approved):
LGTM 👍

jaeyun-jung (approved):

PR #335, opened 2023-05-24
Daemon/DBus: Replace NULL with "" for the DBus out parameters

wooksong:
When the out parameter type is set to "s" (which means a string type value), it should not be a NULL value. This patch fixes such a case in the DBus callback implementation for the pipeline.


myungjoo (approved):

anyj0527 (approved):
LGTM✔

jaeyun-jung (approved):

PR #342, opened 2023-06-13
[fix] Fix mem leak in pipeline API @open sesame 6/14 14:43

anyj0527:
- Free GstTensorsConfig using gst_tensors_config_free
- Initialize ml_tensors_info_s before set a new one


c/src/ml-api-inference-pipeline.c
```diff
@@ -275,6 +275,8 @@ get_tensors_info_from_caps (GstCaps * caps, ml_tensors_info_s * info,
     }
   }
 
+  gst_tensors_config_free (&config);
```
jaeyun-jung, line 278: config at line 260 is not initialized. This may cause unexpected exception.

jaeyun-jung: I recommend adding config-free inside of for(;;) block.

jaeyun-jung (approved):

gichan-jang (approved):
LGTM

PR #346, opened 2023-06-19
Daemon/Meson: Add a pkg-config file for ml-agentd

wooksong:
This PR adds a pkg-config file that provides details for using the Machine Machine Learning Agent service interfaces and includes it in the Tizen package, libmachine-learning-agent-devel.

### DIFF
--- a/daemon/meson.build
+++ b/daemon/meson.build
@@ -141,3 +141,10 @@ configure_file(input: '../dbus/machine-learning-agent.service.in',
   install_dir: systemd_service_dir,
   configuration: systemd_conf
 )
+
+ml_agentd_conf = configuration_data()
+ml_agentd_conf.merge_from(api_conf)
+configure_file(input: 'ml-agentd.pc.in', output: 'ml-agentd.pc',
+  install_dir: join_paths(api_install_libdir, 'pkgconfig'),
+  configuration: ml_agentd_conf
+)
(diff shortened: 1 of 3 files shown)


daemon/meson.build
```diff
@@ -141,3 +141,10 @@ configure_file(input: '../dbus/machine-learning-agent.service.in',
   install_dir: systemd_service_dir,
   configuration: systemd_conf
 )
+
+ml_agentd_conf = configuration_data()
+ml_agentd_conf.merge_from(api_conf)
+configure_file(input: 'ml-agentd.pc.in', output: 'ml-agentd.pc',
```
jaeyun-jung, line 147: Q: dependency to ml-agentd
https://github.com/nnstreamer/api/blob/9307dc7fbb5f79ed9d54a6e4555b275046467f65/c/meson.build#L60
.spec and debian-control
should we add dependency to ml-agentd for ml-service?

wooksong: In the spec file, dependency on ```libmachine-learning-agent``` is declared as follows:
```Requires:       libmachine-learning-agent  = %{version}-%{release}```

If you are asking about the dependency on the service, ```machine-learning-agent```, runtime dependency is required (but this will bring circular dependency) but not for build time dependency.

Runtime dependency will be addressed in another PR after testing some **Weak Dependency** policies on Debian and Tizen.

jaeyun-jung (approved):

songgot (approved):
LGTM, 
Duplicate 'Machine' in commit message

PR #349, opened 2023-06-27
Daemon/Model: Add missing gst initialization

wooksong:
This patch adds GStreamer initialization procedure into init_model_module. Without the initialization, D-Bus interfaces for the models inproperly work.


jaeyun-jung (approved):
LGTM, plz consider adding common func for ml-agent as we talked on offline.

PR #353, opened 2023-07-06
[lts][test] Remove a TC with incompatible TFLite version

anyj0527:
- With tflite 2.8.1, the TC does not behave like as in 2.11.0+, remove it.


again4you (approved):
Verified!

PR #354, opened 2023-07-07
[API] function to get version

jaeyun-jung:
Add util function to get version of ML-API.


c/src/ml-api-common.c
```diff
@@ -16,8 +16,33 @@
...
+  if (major)
+    *major = VERSION_MAJOR;
+  if (minor)
+    *minor = VERSION_MINOR;
+  if (micro)
+    *micro = VERSION_MICRO;
+}
+
+/**
+ * @brief Gets the version string of machine-learning API.
```
songgot, line 38: How about adding  "The returned string should be freed with g_free() when no longer needed." ?

jaeyun-jung: I added it in header, not .c.

songgot: Thank you. I got it.

songgot (approved):
LGTM :+1:

gichan-jang (approved):
LGTM

PR #356, opened 2023-07-13
[option] Use ml_option_get to get value from ml_option_h

anyj0527:
- Use ml_option API rather than accessing internal structure.
- Declare ml_info and related enum and structures. And add internal functions for create, destroy, set, get.
- Replace ml_option_s with the new internal structure in ml_option API.

### DIFF
--- a/c/src/ml-api-common.c
+++ b/c/src/ml-api-common.c
@@ -1287,11 +1287,11 @@ ml_strerror (int errnum)
  * @brief Internal function for destroy value of option table
  */
 static void
-_ml_option_value_free (gpointer data)
+_ml_info_value_free (gpointer data)
 {
-  ml_option_value_s *_value;
+  ml_info_value_s *_value;
 
-  _value = (ml_option_value_s *) data;
+  _value = (ml_info_value_s *) data;
   if (_value) {
     if (_value->destroy)
       _value->destroy (_value->value);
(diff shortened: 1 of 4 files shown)


c/src/ml-api-service-query-client.c
```diff
@@ -98,45 +94,42 @@ ml_service_query_create (ml_option_h option, ml_service_h * h)
...
+  if (ML_ERROR_NONE == ml_option_get (option, "timeout", &value))
+    g_string_append_printf (tensor_query_client_prop, " timeout=%u ",
+        *((guint *) value));
+
+  if (ML_ERROR_NONE != ml_option_get (option, "caps", &value)) {
     g_string_free (tensor_query_client_prop, TRUE);
     _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
         "The option 'caps' must be set before call ml_service_query_create.");
   }
+  caps = g_strdup ((gchar *) value);
```
jaeyun-jung, line 132: inside of the block? line 127

anyj0527: L127 has different expression with other if lines, cause `caps` is necessary for this API.

c/src/ml-api-common.c
```diff
@@ -1287,25 +1287,117 @@ ml_strerror (int errnum)
...
+{
+  ml_info_value_s *info_value;
+
+  if (!info || !key || !value)
+    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
+        "The parameter, 'info', 'key' or 'value' is NULL. It should be a valid ml_info, key and value.");
+
+  info_value = g_hash_table_lookup (info->table, key);
+  if (!info_value)
+    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
```
jaeyun-jung, line 1386: Let's reduce the error message, sometimes business division complains unnecessary error logs.
```if (!info_value) return ML_ERROR_INVALID_PARAMETER;```

c/src/ml-api-service-query-client.c
jaeyun-jung: understood

songgot (approved):
LGTM :+1:

c/src/ml-api-common.c
anyj0527: Thanks, I updated the code PTAL.

jaeyun-jung (approved):

PR #358, opened 2023-07-17
[C-Api] util to check file path

jaeyun-jung:
Code clean, add internal function to check the path of model or resource. This will be used resource-add function later.


anyj0527 (approved):
LGTM✔

songgot (approved):
LGTM :+1:

PR #360, opened 2023-07-17
[Service] util to init gdbus module

jaeyun-jung:
Fix log message, and add common function to initialize gdbus module.

### DIFF
--- a/daemon/gdbus-util.c
+++ b/daemon/gdbus-util.c
@@ -26,7 +26,7 @@ int
 gdbus_export_interface (gpointer instance, const char *obj_path)
 {
   if (g_dbus_sys_conn == NULL) {
-    _E ("cannot get the dbus connection to the system message bus\n");
+    _E ("Cannot get the dbus connection to the system message bus");
     return -ENOSYS;
   }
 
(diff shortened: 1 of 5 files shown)


songgot (approved):
LGTM :+1:

daemon/gdbus-util.c
```diff
@@ -141,3 +141,18 @@ gdbus_put_system_connection (void)
...
   g_clear_object (&g_dbus_sys_conn);
 }
+
+/**
+ * @brief Common function to initailize the DBus module.
+ */
+void
+gdbus_initialize (void)
+{
+  GError *err = NULL;
```
anyj0527, line 151: How about `g_autoptr (GError) = NULL;` for more short code?

daemon/gdbus-util.c
```diff
@@ -141,3 +141,18 @@ gdbus_put_system_connection (void)
 {
   g_clear_object (&g_dbus_sys_conn);
 }
+
+/**
+ * @brief Common function to initailize the DBus module.
```
anyj0527, line 146: minor typo `initailize` here and in gdbus-util.h

anyj0527 (approved):
LGTM with some insignificant comments

daemon/gdbus-util.c
jaeyun-jung: updated

daemon/gdbus-util.c
jaeyun-jung: For future work, I am not sure another code for initializing ml-agent module which requires gerror ptr. Let's keep this in this time.

PR #361, opened 2023-07-18
[Service/CodeClean] reduce duplicated code

jaeyun-jung:
Code clean,
1. Add methods in service-db to reduce duplicated code.
2. Add key prefix for each table.


songgot (approved):
Looks good to me!

PR #364, opened 2023-07-28
[Service] data from ml-service

jaeyun-jung:
1. Get resource info with insertion order.
2. We do not neet to export key from service-db. 
    Add common json info to get the data from db and hide key from json info.


anyj0527 (approved):
LGTM✔

PR #367, opened 2023-07-31
[Service] return val to parse app info

jaeyun-jung:
Code clean, return val if failed to parse app-info from json string.


anyj0527 (approved):
LGTM ✔

PR #368, opened 2023-07-31
[Service] util to get app info

jaeyun-jung:
Define internal common function to get app info.

TODO:
New field for app info for ml-resource and the funtion adding ml-resource will be udpated.


anyj0527 (approved):
LGTM✔

PR #370, opened 2023-07-31
[Service] handle app-info for ml-resource

jaeyun-jung:
Add new field app_info for tizen RPK install case, to manage app-information like model API.


anyj0527 (approved):
LGTM✔

PR #371, opened 2023-07-31
[Service] util to parse app-info

jaeyun-jung:
Use common function to parse app-info and update ml-information handle.


songgot (approved):
LGTM!!

PR #372, opened 2023-08-01
[Service/Test] valid ml-information handle

jaeyun-jung:
Set valid handle name in testcase - ml-information handle for model and resource API.


gichan-jang (approved):
LGTM

PR #374, opened 2023-08-02
[Pipeline] element name for tizen media src

jaeyun-jung:
Code clean, get the element name from tizen media config and replace it.

TODO: handle properties for each element.


songgot (approved):
Looks good to me.

PR #375, opened 2023-08-03
[CodeClean/Service] util to get app-info

jaeyun-jung:
Code clean, add util to get app-info string and reduce duplicated code.


gichan-jang (approved):
LGTM

PR #376, opened 2023-08-03
[CodeClean/Service] parse json in rpk

jaeyun-jung:
Code clean, define function to parse json file in rpk, including resource and pipeline update.

### DIFF
--- a/daemon/pkg-mgr.cc
+++ b/daemon/pkg-mgr.cc
@@ -55,6 +55,92 @@ _get_app_info (const gchar *package_name, const gchar *res_type, const gchar *re
 }
 
 /**
+ * @brief Parse model json and update ml-service database.
+ */
+static void
+_parse_model_json (const gchar *json_path, const gchar *app_info)
+{
+  if (!g_file_test (json_path, (GFileTest) (G_FILE_TEST_EXISTS | G_FILE_TEST_IS_REGULAR))) {
+    _W ("Failed to find json file '%s'. RPK using ML Service API should provide this json file.",
+        json_path);
+    return;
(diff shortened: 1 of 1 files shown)


daemon/pkg-mgr.cc
```diff
@@ -54,6 +54,92 @@ _get_app_info (const gchar *package_name, const gchar *res_type, const gchar *re
   return json_generator_to_data (gen, NULL);
 }
 
+/**
+ * @brief Parse model json and update ml-service database.
+ */
+static void
+_parse_model_json (const gchar *json_path, const gchar *app_info)
```
songgot, line 61: Is app_info nullable?

jaeyun-jung: internally, impossible.

songgot: Thank you!!

daemon/pkg-mgr.cc
```diff
@@ -159,66 +245,16 @@ _pkg_mgr_event_cb (const char *type, const char *package_name,
...
-        db.set_model (name, model, true, description, app_info, &version);
-        _I ("The model with app_info (%s) is registered as version %u", app_info, version);
-      }
-    } catch (const std::exception &e) {
-      _E ("%s", e.what ());
-    }
-
-    db.disconnectDB ();
-
+    g_autofree gchar *pipeline_json
```
myungjoo, line 250: glib has an API for merging paths. Use https://docs.gtk.org/glib/func.build_filename.html

myungjoo (approved):

PR #380, opened 2023-08-08
[C-Api] check index in tensors-info handle

jaeyun-jung:
Code clean, check index in tensors-info handle and update log message.


songgot (approved):
LGTM!

wooksong (approved):

PR #395, opened 2023-09-12
[Service] Check return value

gichan-jang:
- Fix unchecked return value.

### DIFF
--- a/c/src/ml-api-remote-service.c
+++ b/c/src/ml-api-remote-service.c
@@ -218,13 +218,26 @@ _mlrs_model_register (gchar * service_key, nns_edge_data_h data_h,
   GError *error = NULL;
   gboolean ret = TRUE;
 
-  nns_edge_data_get_info (data_h, "description", &description);
-  nns_edge_data_get_info (data_h, "name", &name);
-  nns_edge_data_get_info (data_h, "activate", &activate);
+  if (NNS_EDGE_ERROR_NONE != nns_edge_data_get_info (data_h, "description",
+          &description)
+      || NNS_EDGE_ERROR_NONE != nns_edge_data_get_info (data_h, "name", &name)
+      || NNS_EDGE_ERROR_NONE != nns_edge_data_get_info (data_h, "activate",
+          &activate)) {
+    _ml_loge ("Failed to get info from data handle.");
+    ret = FALSE;
+    goto error;
+  }
+
   active_bool = _mlrs_parse_activate (activate);
 
   dir_path = g_build_path ("/", current_dir, service_key, NULL);
-  g_mkdir_with_parents (dir_path, 0755);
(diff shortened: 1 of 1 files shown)


c/src/ml-api-remote-service.c
```diff
@@ -218,13 +218,27 @@ _mlrs_model_register (gchar * service_key, nns_edge_data_h data_h,
...
+    goto error;
+  }
+
   active_bool = _mlrs_parse_activate (activate);
 
   dir_path = g_build_path ("/", current_dir, service_key, NULL);
-  g_mkdir_with_parents (dir_path, 0755);
+  if (g_mkdir_with_parents (dir_path, 0755) < 0) {
+    _ml_loge ("Failed to create directory %s., error: %s", dir_path,
+        error ? error->message : "unknown error");
```
jaeyun-jung, line 236: invalid param error (it is not set)

gichan-jang: Oh, Thanks. Updated.

c/src/ml-api-remote-service.c
```diff
@@ -562,28 +584,55 @@ ml_remote_service_register (ml_service_h handle, ml_option_h option, void *data,
...
+  }
+  ret = nns_edge_data_set_info (data_h, "service-key", service_key);
+  if (NNS_EDGE_ERROR_NONE != ret) {
+    _ml_error_report ("Failed to set service key in edge data.");
+    goto done;
+  }
+  ret = ml_option_get (option, "description", (void **) &description);
+  if (ML_ERROR_NONE != ret) {
+    _ml_logi ("Failed to get option description.");
+  }
```
songgot, line 600: Does it need ```goto done``` ?

c/src/ml-api-remote-service.c
```diff
@@ -218,13 +218,26 @@ _mlrs_model_register (gchar * service_key, nns_edge_data_h data_h,
...
+    _ml_loge ("Failed to get info from data handle.");
+    ret = FALSE;
+    goto error;
+  }
+
   active_bool = _mlrs_parse_activate (activate);
 
   dir_path = g_build_path ("/", current_dir, service_key, NULL);
-  g_mkdir_with_parents (dir_path, 0755);
+  if (g_mkdir_with_parents (dir_path, 0755) < 0) {
```
wooksong, line 234: Before making a new directory, is it necessary to check that the target directory exists?

gichan-jang: If the directory already exists, this returns 0. So I think don't need to check the the existence of the directory.

c/src/ml-api-remote-service.c
gichan-jang: It's not necessary option, so just inform to the user.

wooksong (approved):
LGTM

myungjoo (approved):

PR #402, opened 2023-10-04
[C-Api/Query] use caps property

jaeyun-jung:
Use 'caps' property in appsrc, not caps-filter.

### DIFF
--- a/c/src/ml-api-service-query-client.c
+++ b/c/src/ml-api-service-query-client.c
@@ -134,7 +134,7 @@ ml_service_query_create (ml_option_h option, ml_service_h * h)
   prop = g_string_free (tensor_query_client_prop, FALSE);
   description =
       g_strdup_printf
-      ("appsrc name=srcx ! %s ! tensor_query_client %s name=qcx ! tensor_sink name=sinkx async=false sync=false",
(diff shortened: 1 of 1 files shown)


c/src/ml-api-service-query-client.c
```diff
@@ -134,7 +134,7 @@ ml_service_query_create (ml_option_h option, ml_service_h * h)
   prop = g_string_free (tensor_query_client_prop, FALSE);
   description =
       g_strdup_printf
-      ("appsrc name=srcx ! %s ! tensor_query_client %s name=qcx ! tensor_sink name=sinkx async=false sync=false",
+      ("appsrc name=srcx caps=%s ! tensor_query_client %s name=qcx ! tensor_sink name=sinkx async=false sync=false",
```
gichan-jang, line 137: Why do you change this?

jaeyun-jung: I recommend using caps property when using appsrc, to clearly denote the caps from appsrc. caps-filter may not pass the caps to appsrc in some case.

gichan-jang (approved):
LGTM

anyj0527 (approved):
LGTM✔

PR #410, opened 2023-10-18
[ml-service] Set transaction

gichan-jang:
There is some timing issue to save the changes in DB, set trasnsation for INSERT/UPDATE command.


anyj0527 (approved):
LGTM✔

jaeyun-jung (approved):

PR #412, opened 2023-10-24
[ml-service] Set nns-edge handle ID.

gichan-jang:
Let's generate edge handle ID automatically by default.
The user also set edge ID using ml_option_set().

Signed-off-by: gichan2-jang <gichan2.jang@samsung.com>

### DIFF
--- a/c/src/ml-api-service-remote.c
+++ b/c/src/ml-api-service-remote.c
@@ -51,6 +51,7 @@ typedef struct
   guint dest_port;
   nns_edge_connect_type_e conn_type;
   nns_edge_node_type_e node_type;
+  gchar *id;
 } edge_info_s;
 
 /**
@@ -138,6 +139,10 @@ _mlrs_get_edge_info (ml_option_h option, edge_info_s * edge_info)
     edge_info->topic = g_strdup (value);
   if (ML_ERROR_NONE == ml_option_get (option, "node-type", &value))
     edge_info->node_type = _mlrs_get_node_type (value);
+  if (ML_ERROR_NONE == ml_option_get (option, "id", &value)) {
+    g_free (edge_info->id);
+    edge_info->id = g_strdup (value);
+  }
 }
 
 /**
@@ -169,6 +174,7 @@ _mlrs_release_edge_info (edge_info_s * edge_info)
   g_free (edge_info->dest_host);
   g_free (edge_info->host);
   g_free (edge_info->topic);
+  g_free (edge_info->id);
 }
 
 /**
@@ -503,7 +509,7 @@ _mlrs_create_edge_handle (_ml_remote_service_s * remote_s,
   int ret = 0;
   nns_edge_h edge_h = NULL;
 
-  ret = nns_edge_create_handle (edge_info->topic, edge_info->conn_type,
+  ret = nns_edge_create_handle (edge_info->id, edge_info->conn_type,
       edge_info->node_type, &edge_h);
 
   if (NNS_EDGE_ERROR_NONE != ret) {
@@ -595,6 +601,7 @@ ml_service_remote_create (ml_option_h option, ml_service_event_cb cb,
   edge_info->dest_host = g_strdup ("localhost");
   edge_info->dest_port = 0;
   edge_info->conn_type = NNS_EDGE_CONNECT_TYPE_UNKNOWN;
+  edge_info->id = NULL;
 
   _mlrs_get_edge_info (option, edge_info);
 


c/src/ml-api-service-remote.c
```diff
@@ -451,41 +495,45 @@ _mlrs_edge_event_cb (nns_edge_event_h event_h, void *user_data)
...
     _ml_error_report ("nns_edge_start failed.");
-    nns_edge_release_handle (*edge_h);
+    nns_edge_release_handle (edge_h);
     return ret;
   }
 
   if (edge_info->node_type == NNS_EDGE_NODE_TYPE_SUB) {
-    ret = nns_edge_connect (*edge_h, edge_info->dest_host,
-        edge_info->dest_port);
+    ret = nns_edge_connect (edge_h, edge_info->dest_host, edge_info->dest_port);
```
wooksong, line 529: I am just wondering. It looks that ```edge_h``` already contains edge_info. Then, Why are those other parameters for the host and port necessary?

c/src/ml-api-service-remote.c
```diff
@@ -280,17 +278,45 @@ _mlrs_model_register (gchar * service_key, nns_edge_data_h data_h,
...
+
+  if (!g_file_test (dir_path, G_FILE_TEST_IS_DIR)) {
+    _ml_loge
+        ("The given param, dir path = \"%s\" is invalid or the dir is not found or accessible.",
+        dir_path);
+    g_free (dir_path);
+    return NULL;
+  }
+
+  return dir_path;
```
wooksong, line 317:
```suggestion
  g_autofree gchar *dir_path = NULL;

  if (remote_s->path) {
    dir_path = g_strdup (remote_s->path);
  } else {
    g_autofree gchar *current_dir = g_get_current_dir ();

    dir_path = g_build_path ("/", current_dir, service_key, NULL);
    if (g_mkdir_with_parents (dir_path, 0755) < 0) {
      _ml_loge ("Failed to create directory %s., error: %s", dir_path,
          g_strerror (errno));
      return NULL;
    }
  }

  if (!g_file_test (dir_path, G_FILE_TEST_IS_DIR)) {
    _ml_loge
        ("The given param, dir path = \"%s\" is invalid or the dir is not found or accessible.",
        dir_path);
    return NULL;
  }

  return g_steal_pointer(dir_path);
```

c/src/ml-api-service-remote.c
gichan-jang: Right, since we set edge info, we don't need dest host and port as you said. This is duplicate work.
However, we will not change this because the user may use the connection without the set info.

c/src/ml-api-service-remote.c
gichan-jang: Thanks, updated

c/src/ml-api-service-remote.c
```diff
@@ -366,27 +394,35 @@ _mlrs_process_remote_service (nns_edge_data_h data_h)
...
+      g_autofree gchar *dir_path = NULL;
 
       if (!_mlrs_get_data_from_uri ((gchar *) data, array)) {
         g_byte_array_free (array, TRUE);
         _ml_error_report_return (NNS_EDGE_ERROR_IO,
             "Failed to get data from uri: %s.", (gchar *) data);
       }
-      if (!_mlrs_model_register (service_key, data_h, array->data, array->len)) {
+
+      dir_path = _mlrs_get_model_dir_path (remote_s, service_key);
```
songgot, line 405: _mlrs_get_model_dir_path() may return null in an ERROR.
How about returning an error if dir_path is NULL?

c/src/ml-api-service-remote.c
```diff
@@ -366,27 +394,35 @@ _mlrs_process_remote_service (nns_edge_data_h data_h)
...
         ret = NNS_EDGE_ERROR_UNKNOWN;
       }
       g_byte_array_free (array, TRUE);
+      event_type = ML_SERVICE_EVENT_MODEL_REGISTERED;
       break;
     }
     case ML_REMOTE_SERVICE_TYPE_MODEL_RAW:
     {
-      if (!_mlrs_model_register (service_key, data_h, data, data_len)) {
+      g_autofree gchar *dir_path =
```
songgot, line 418: same

c/src/ml-api-service-remote.c
```diff
@@ -451,41 +495,45 @@ _mlrs_edge_event_cb (nns_edge_event_h event_h, void *user_data)
...
+  ret = nns_edge_create_handle (edge_info->id, edge_info->conn_type,
+      edge_info->node_type, &edge_h);
 
   if (NNS_EDGE_ERROR_NONE != ret) {
     _ml_error_report ("nns_edge_create_handle failed.");
     return ret;
   }
 
-  ret = nns_edge_set_event_callback (*edge_h, _mlrs_edge_event_cb, NULL);
+  ret = nns_edge_set_event_callback (edge_h, _mlrs_edge_event_cb, remote_s);
```
songgot, line 512: Internally, remote_s seems to always be non-null.

c/src/ml-api-service-remote.c
gichan-jang: This was resolved in previous PR. Thanks.

jaeyun-jung (approved):

myungjoo (approved):
LGTM.

PR #413, opened 2023-10-27
[Service] active flag from json

jaeyun-jung:
1. Update 'activate' flag from json when registering new model info from res-pkg.
2. Allow null description in json, description string is not mandatory field.
3. Fix log message, hide app info string.


songgot (approved):
LGTM

PR #416, opened 2023-11-14
[C-Api/Pipeline] create handle before calling sink callback

jaeyun-jung:
1. Create tensors-info handle before calling sink callback.
2. We do not need to use ml-info in pipeline elem struct. Use gst-info instead.


anyj0527 (approved):
LGTM👍

PR #423, opened 2023-11-17
[Common/CodeClean] struct for tensor memories

jaeyun-jung:
Code clean, use GstTensorMemory struct defined in nnstreamer.


anyj0527 (approved):
LGTM👍

PR #424, opened 2023-11-21
[C-Api] use tensor-info struct from nns

jaeyun-jung:
Use gst-info struct and util functions from nnstreamer.
Remove unnecessary functions for ml-info.


anyj0527 (approved):
LGTM👍 Great!

myungjoo (approved):

PR #427, opened 2023-12-05
[C-Api/Common] check invalid dimension

jaeyun-jung:
Check invalid dimension case before updating tensor-info handle.


songgot (approved):
LGTM

myungjoo (approved):

PR #435, opened 2024-01-04
[CodeClean] dependency to ml-agent

jaeyun-jung:
Prepare deleting ml-agent from api repo, set dependency to ml-agent and remove unnecessary line.


anyj0527 (approved):
LGTM

PR #444, opened 2024-01-15
[C-Api/Single] data handle passed to invoke thread

jaeyun-jung:
Set in/out data handle those are passed to invoke thread, also clearly release data when closing the thread.


anyj0527 (approved):
LGTM✔

PR #445, opened 2024-01-15
[Test] remove unnecessary info handle

jaeyun-jung:
In testcase of ArmNN, remove unnecessary tensor-info and compare dimension after getting info handle from model.


anyj0527 (approved):
LGTM✔

myungjoo (approved):

PR #447, opened 2024-01-18
[Dist/Tizen] Add omitted build dependency on libtzplatform-config

wooksong:
This patch is a trivial bug fix that adds an omitted build dependency on libtzplatform-config to the Tizen RPM spec file.

CC: @jaeyun-jung 


jaeyun-jung (approved):

myungjoo:
If this is required "BuildRequires", how could CI build the package w/ gbs without it?

myungjoo (approved):

myungjoo:
Need a checker that ensures proper DB path setting at build/packaging time.

PR #452, opened 2024-01-30
[CodeClean] fix svace issue

jaeyun-jung:
Code clean, fix svace issues about lock in single/pipeline handle.

### DIFF
--- a/c/src/ml-api-inference-pipeline.c
+++ b/c/src/ml-api-inference-pipeline.c
@@ -1776,15 +1776,11 @@ static ml_pipeline_src_callbacks_s *
 get_app_src_callback (ml_pipeline_common_elem * src_h, void **data)
 {
   ml_pipeline_src_callbacks_s *src_cb = NULL;
-  ml_pipeline_element *elem;
 
-  elem = src_h->element;
-  g_mutex_lock (&elem->lock);
   if (src_h->callback_info) {
     src_cb = &src_h->callback_info->src_cb;
     *data = src_h->callback_info->pdata;
   }
-  g_mutex_unlock (&elem->lock);
 
   return src_cb;
 }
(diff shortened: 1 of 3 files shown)


again4you (approved):
LGTM!

c/src/ml-api-inference-pipeline.c
```diff
@@ -1777,14 +1777,21 @@ get_app_src_callback (ml_pipeline_common_elem * src_h, void **data)
 {
   ml_pipeline_src_callbacks_s *src_cb = NULL;
   ml_pipeline_element *elem;
+  ml_pipeline *p;
 
+  p = src_h->pipe;
   elem = src_h->element;
+
+  g_mutex_lock (&p->lock);
```
myungjoo, line 1785: What's the op/data protected by this lock in line 1788-1791?

jaeyun-jung: This function returns appsrc callback which sets with pipeline lock, so svace-checker returns error when getting cb info without p-lock.

niley7464 (approved):
LGTM!

c/src/ml-api-inference-pipeline.c
myungjoo: The problem is that... ```ml_pipeline_common_elem``` has ```{ callback_info_s , ml_pipeline }```
and ```ml_pipeline->lock``` is protecting ```callback_info_s->*```.

this lock should protect elements in ml_pipeline, not others in ml_pipeline_common_elem.
If it really needs to protect  callback_info_s, it means the lock is really messed up.

myungjoo: The lock needs to declare (in doxygen) which are protected by the lock.

songgot (approved):
LGTM

myungjoo (approved):

PR #458, opened 2024-02-20
[Service] condition to validate handle

jaeyun-jung:
Check priv data when validating ml-service handle.


c/src/ml-api-service-common.c
```diff
@@ -28,6 +28,18 @@ _ml_service_handle_is_valid (ml_service_s * mls)
   if (mls->magic != ML_SERVICE_MAGIC)
     return FALSE;
 
+  switch (mls->type) {
+    case ML_SERVICE_TYPE_SERVER_PIPELINE:
+    case ML_SERVICE_TYPE_CLIENT_QUERY:
+    case ML_SERVICE_TYPE_REMOTE:
```
gichan-jang, line 34: Need to check `ML_SERVICE_TYPE_EXTENSION`

gichan-jang (approved):
LGTM

PR #459, opened 2024-02-21
[CodeClean/Service] set null after free priv data

jaeyun-jung:
1. set null after releasing priv data in service handle.
2. remove unnecessary param in query priv data.


gichan-jang (approved):
LGTM

PR #460, opened 2024-02-27
[ml-service] Integrates remote service into ml-service

gichan-jang:
Integrates remote service into ml-service.

@todo: File name and source location will be changed later.

Signed-off-by: gichan2-jang <gichan2.jang@samsung.com>


c/src/ml-api-service-remote.c
```diff
@@ -756,3 +754,41 @@ ml_service_remote_register (ml_service_h handle, ml_option_h option, void *data,
...
+
+  remote_s = (_ml_remote_service_s *) mls->priv;
+
+  status = ml_option_set (remote_s->info, name, value, destroy);
+  if (status != ML_ERROR_NONE) {
+    _ml_error_report_return (status, "Failed to set the information in ml-service remote handle.");
+  }
+
+  if (g_ascii_strcasecmp (name, "path") == 0) {
+    status = ml_service_remote_set_path (mls);
```
niley7464, line 793: This function only sets path now. Is there any plan to support other names in the future?
I think it would be better make error or log to this part if we don't have any plan to support other options.

Also although this is not for this PR, it seems necessary to write a test case for `ml_service_(remote_)set_information`

c/src/ml-api-service-remote.c
```diff
@@ -563,81 +565,77 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
-  }
+  mls = (ml_service_s *) handle;
 
   mls->priv = remote_s = g_try_new0 (_ml_remote_service_s, 1);
   if (remote_s == NULL) {
-    _ml_service_destroy_internal (mls);
     _ml_error_report_return (ML_ERROR_OUT_OF_MEMORY,
         "Failed to allocate memory for the service handle's private data. Out of memory?");
   }
+  remote_s->info = option;
```
niley7464, line 627: If we just use `ml_option_h option`, can I know what happens if the option changes or release outside of the service after?
If unintended changes can occur, how about add a function like `ml_option_copy or clone()`?

niley7464 (commented):
Thank you for your hard work integrating existing functions! 👍 👍 
I added some comments for this PR.

I also have an opinion on the file structure.
How about separate ml-api-service-extension.c to ml-api-service.c and ml-api-service-extension.c ? 
ml-api-service.c : `ml_service_new` ,`ml_service_start` 
ml-api-service-extension.c : `_ml_extension_node_info_new`, `_ml_extension_node_info_free`

It will make it easier to add features and improve readability.

c/src/ml-api-service-extension.c
```diff
@@ -1236,13 +1332,54 @@ ml_service_get_information (ml_service_h handle, const char *name, char **value)
...
   }
 
   return status;
 }
 
+/**
+ * @brief Sets the information in ml-service handle.
+ */
+int
+ml_service_set_information (ml_service_h handle, const char *name, void *value,
```
jaeyun-jung, line 1347: Discussion: set/get string-type information 
ml_service_set_information (ml_service_h handle, const char *name, const char *value);

c/src/ml-api-service-extension.c
```diff
@@ -1324,3 +1461,70 @@ ml_service_destroy (ml_service_h handle)
 
   return _ml_service_destroy_internal ((ml_service_s *) handle);
 }
+
+/**
+ * @brief Request the service related to machine-learning.
+ */
+int
+ml_service_request (ml_service_h handle, const char *config_path,
```
jaeyun-jung, line 1469: Suggestion: same param with ml_service_add_input() 
ml_service_request (ml_service_h handle, const char *name, const ml_tensors_data_h data);

I think we can use common function for request.
case 1. use ml_service_set_information()
case 2. add param -json string (e.g., ml_service_request (ml_service_h handle, const char *name, const ml_tensors_data_h data, const char *json_option);)

c/src/ml-api-service-extension.c
```diff
@@ -723,6 +728,46 @@ _ml_extension_conf_parse_json (ml_service_s * mls, const char *config_path)
...
 }
 
+/**
+ * @brief Internal function to create ml-service extension handle.
+ */
+static int
+_ml_service_create_extension (ml_extension_s ** ext)
+{
+  *ext = g_try_new0 (ml_extension_s, 1);
+  if (ext == NULL) {
```
songgot, line 739: ``` if (*ext == NULL) ``` ?

c/src/ml-api-service-extension.c
```diff
@@ -1324,3 +1461,70 @@ ml_service_destroy (ml_service_h handle)
...
+
+  if (!data)
+    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
+        "The parameter, data (ml_tensors_data_h), is NULL. It should be a valid ml_tensor_data_h instance, which is usually created by ml_tensors_data_create().");
+
+  switch (mls->type) {
+    case ML_SERVICE_TYPE_EXTENSION:
+      break;
+    case ML_SERVICE_TYPE_REMOTE:
+    {
```
songgot, line 1494: How about checking the type ```mls->type == ML_SERVICE_TYPE_REMOTE``` when entering a function instead of the Switch statement?

c/src/ml-api-service-extension.c
```diff
@@ -829,14 +835,89 @@ ml_service_extension_destroy_internal (ml_service_s * mls)
...
+  status = ml_option_set (option, "node-type", node_type, g_free);
+  if (status != ML_ERROR_NONE) {
+    ml_option_destroy (option);
+    _ml_error_report_return (status,
+        "Cannot parse config file. Failed to set node type");
+  }
+
+  status = _ml_service_remote_conf_to_opt (object, node_type, option);
+  if (status != ML_ERROR_NONE) {
+    ml_option_destroy (option);
```
songgot, line 901: If the condition is true, option becomes NULL.  so option cannot be destroyed.

c/src/ml-api-service-extension.c
```diff
@@ -829,14 +835,89 @@ ml_service_extension_destroy_internal (ml_service_s * mls)
...
+  remote_service_object = json_object_get_object_member (object, name);
+
+  list = json_object_get_members (remote_service_object);
+  for (iter = list; iter != NULL; iter = g_list_next (iter)) {
+    val = json_object_get_string_member (remote_service_object, iter->data);
+    status = ml_option_set (option, iter->data, g_strdup (val), g_free);
+    if (status != ML_ERROR_NONE) {
+      _ml_error_report ("Failed to set %s option: %s.", (gchar *) list->data,
+          val);
+      option = NULL;
```
songgot, line 859: Is it necessary?

c/src/ml-api-service-remote.c
```diff
@@ -563,81 +564,78 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
+  return ML_ERROR_NONE;
+}
+
 /**
  * @brief Creates ml-service handle with given ml-option handle.
  */
 int
-ml_service_remote_create (ml_option_h option, ml_service_event_cb cb,
-    void *user_data, ml_service_h * handle)
+ml_service_remote_create (ml_service_h handle, ml_option_h option)
```
songgot, line 605: How about checking if option is NULL?

c/src/ml-api-service-remote.c
```diff
@@ -563,81 +564,78 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
+      _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
+          "The given param, dir path = \"%s\" is invalid or the dir is not found or accessible.",
+          _path);
+    }
+    if (g_access (_path, W_OK) != 0) {
+      _ml_error_report_return (ML_ERROR_PERMISSION_DENIED,
+          "Write permission denied, path: %s", _path);
+    }
+  }
+  mlrs->path = g_strdup (_path);
```
songgot, line 596: How about moving this line to 595?

tests/test_models/config/remote_service_model_raw_register.conf
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
gichan-jang: Thanks for the review. I considered whether the configs could be combined or not, but it was difficult to combine because ml-service type is determined by the initial setting value such as a sender or receiver.

tests/capi/unittest_capi_remote_service.cc
```diff
@@ -135,87 +136,51 @@ _ml_service_event_cb (ml_service_event_e event_type, void *user_data)
...
+ */
+TEST_F (MLRemoteService, registerPipeline)
+{
+  int status;
 
-  status = ml_service_remote_create (client_option_h, NULL, NULL, &client_h);
-  EXPECT_EQ (ML_ERROR_NONE, status);
+  ml_service_h client_h;
+  g_autofree gchar *sender_config = _get_config_path ("remote_service_sender.conf");
+  status = ml_service_new (sender_config, &client_h);
```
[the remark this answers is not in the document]
gichan-jang: Agree! I applied google test fixture class. 
This method will improve the readability of the test cases and reduce human errors.
Thanks.

c/src/ml-api-service-remote.c
gichan-jang: Thanks. `ml_service_set_information()` was added as you suggested by @jaeyun-jung.
All general information is set by this method, but the only one property of ml-service-remote changed by this method is `path`. Other information has no effect on the ml-service.
If there is additional information that needs to be set in ml-service property, it is easy to modify because it only needs to modify the internal implementation without changing the API.
So, If necessary, let's change it later!

c/src/ml-api-service-remote.c
gichan-jang: As you mentioned, it is dangerous to copy only the pointer of the ml-option.
And instead of copying the whole ml-option, it was changed to save only the necessary information as an property of ml-service.

c/src/ml-api-service-extension.c
gichan-jang: Changed from `void *` to `char *`. Thanks!

c/src/ml-api-service-extension.c
gichan-jang: Nice!
Remove ml_sercice_add_input and changed to use ml_service_request ().

ml_service_request (ml_service_h handle, const char *name, const char *config_path, const ml_tensors_data_h data);

c/src/ml-api-service-extension.c
gichan-jang: Updated.

c/src/ml-api-service-extension.c
gichan-jang: As ml-service-extension was additionally implemented, there was no need to change it. PTAL.

c/src/ml-api-service-extension.c
gichan-jang: Good point. Updated.

c/src/ml-api-service-extension.c
gichan-jang: Removed.

c/src/ml-api-service-remote.c
gichan-jang: Good point. Updated.

c/src/ml-api-service-remote.c
gichan-jang: Updated.

gichan-jang:
Please review changes again.

c/src/ml-api-service.c
```diff
@@ -288,6 +295,129 @@ _ml_service_conf_parse_information (ml_service_s * mls, JsonObject * info)
...
+      status = ml_service_remote_set_service (mls, iter->data, val);
+      if (status!= ML_ERROR_NONE) {
+        _ml_error_report ("Failed to set service key : %s", iter->data);
+        break;
+      }
+    }
+  }
+  g_list_free (list);
+
+  return ML_ERROR_NONE;
```
jaeyun-jung, line 356: return status?

c/src/ml-api-service.c
```diff
@@ -288,6 +295,129 @@ _ml_service_conf_parse_information (ml_service_s * mls, JsonObject * info)
...
+
+  status = ml_option_create (&option);
+  if (status != ML_ERROR_NONE) {
+    _ml_error_report_return (status, "Failed to create ml-option.");
+  }
+
+  status = _ml_service_remote_conf_to_opt (object, "remote", option);
+  if (status != ML_ERROR_NONE) {
+    goto done;
+    _ml_error_report ("Failed to set ml-option from config file.");
```
jaeyun-jung, line 376: log report and goto done

c/src/ml-api-service.c
```diff
@@ -288,6 +295,129 @@ _ml_service_conf_parse_information (ml_service_s * mls, JsonObject * info)
...
+  }
+
+  status = ml_service_remote_create (mls, option);
+  if (status != ML_ERROR_NONE) {
+    _ml_error_report ("Failed to create ml-service-remote.");
```
jaeyun-jung: need error handle (goto done)

c/src/ml-api-service.c
```diff
@@ -344,14 +474,7 @@ ml_service_new (const char *config, ml_service_h * handle)
...
 
-  if (json_object_has_member (object, "single") ||
-      json_object_has_member (object, "pipeline")) {
-    service_type = ML_SERVICE_TYPE_EXTENSION;
-  } else {
-    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
-        "Failed to parse configuration file, cannot get the valid type from configuration.");
-  }
-
+  service_type = _ml_service_get_type (object);
```
jaeyun-jung, line 477: need handle error case if service type is ML_SERVICE_TYPE_UNKNOWN

gichan-jang: Thanks, updated!

c/src/ml-api-service-remote.c
```diff
@@ -563,9 +564,11 @@ ml_service_remote_release_internal (ml_service_s * mls)
 
   if (mlrs->edge_h) {
     nns_edge_release_handle (mlrs->edge_h);
+  }
```
niley7464, line 569: Waiting for release using sleep funciton has a risk of producing different results depending on the environment.
So I think it would be better to delete `g_usleep` like this 👍 

It's just a question, but I wonder if factors were considered that could cause side effects after removal sleep function.

c/src/ml-api-service-remote.c
```diff
@@ -575,70 +578,81 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
+    _ml_error_report
+        ("Failed to allocate memory for the table of ml-service remote. Out of memory?");
+  }
 
-  remote_s->node_type = edge_info->node_type;
-  remote_s->event_cb = cb;
-  remote_s->user_data = user_data;
-  remote_s->path = g_strdup (_path);
+  if (ML_ERROR_NONE == ml_option_get (option, "path", (void **) (&_path))) {
+    ret = ml_service_remote_set_path (mls, _path);
```
niley7464: Readability and stability seem to have improved by making `ml_service_remote_set_path` function 😄

c/src/ml-api-service-remote.c
```diff
@@ -575,70 +578,81 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
-ml_service_remote_create (ml_option_h option, ml_service_event_cb cb,
-    void *user_data, ml_service_h * handle)
+ml_service_remote_create (ml_service_h handle, ml_option_h option)
 {
   ml_service_s *mls;
   _ml_remote_service_s *remote_s = NULL;
   edge_info_s *edge_info = NULL;
   int ret = ML_ERROR_NONE;
-  gchar *_path = NULL;
+  gchar *_path;
```
niley7464, line 618: It seems there is no problem in logic to remove NULL declaration.
Are there any reason to remove this? (Just curious)

c/src/ml-api-service-remote.c
```diff
@@ -133,13 +132,13 @@ _mlrs_get_edge_info (ml_option_h option, edge_info_s ** edge_info)
   else
     _info->host = g_strdup ("localhost");
   if (ML_ERROR_NONE == ml_option_get (option, "port", &value))
-    _info->port = *((guint *) value);
+    _info->port = g_ascii_strtoull (value, NULL, 10);
```
niley7464, line 135: It seems better to use a `g_ascii_strtoull `function than just casting 👍

c/src/ml-api-service-remote.c
```diff
@@ -82,9 +81,9 @@ _mlrs_get_node_type (const gchar * value)
   if (!value)
     return node_type;
 
-  if (g_ascii_strcasecmp (value, "remote_sender") == 0) {
+  if (g_ascii_strcasecmp (value, "sender") == 0) {
     node_type = NNS_EDGE_NODE_TYPE_PUB;
-  } else if (g_ascii_strcasecmp (value, "remote_receiver") == 0) {
+  } else if (g_ascii_strcasecmp (value, "receiver") == 0) {
```
niley7464, line 86: It seems that now we can only set pub/sub when setting node_type.
At the link below, it says that the pub/sub pattern is a type of obsever pattern and is unaware of each other's existence.
https://handhand.tistory.com/271

In the case of training offloading, I think it would be necessary for the receiver to send the trained model back to the sender.
In this situation, if we use the pub/sub structure, we must create a new service.
Since nnstreamer-edge does not have a reply function, it seems difficult to solve this problem in the current situation.

Although it is not a necessary part of this PR, it would be better to consider the following parts in future development.
```
typedef enum {
  NNS_EDGE_NODE_TYPE_QUERY_CLIENT = 0,
  NNS_EDGE_NODE_TYPE_QUERY_SERVER,
  NNS_EDGE_NODE_TYPE_PUB,
  NNS_EDGE_NODE_TYPE_SUB,

  NNS_EDGE_NODE_TYPE_UNKNOWN,
} nns_edge_node_type_e;
```

`nns_edge_node_type_e` has query_client and server. 
If we set node type with query_client as the sender and query_server as the receiver, we can send two-way data transmission and requests.

niley7464 (commented):
Thank you for integrating remote service into ml-service 👏

I left some comments about something trivial and a topic for discussion.
I would appreciate it if you would consider it when you have time.

c/src/ml-api-service.c
```diff
@@ -67,6 +68,12 @@ _ml_service_set_information_internal (ml_service_s * mls, const char *name,
     case ML_SERVICE_TYPE_EXTENSION:
       status = ml_service_extension_set_information (mls, name, value);
       break;
+    case ML_SERVICE_TYPE_REMOTE:
+    {
+      if (g_ascii_strcasecmp (value, "path") == 0)
```
jaeyun-jung, line 73: I recommend adding _set_information() function and comparing 'name' for later work. Now offloading feature handles path only, but we may need another info.

c/src/ml-api-service-remote.c
gichan-jang: I think there is no need for sleep time because edge_h waits until the thread is destroyed using the join.
The use of sleep time seems an excessive defense code in the initial design.
Thanks.

c/src/ml-api-service-remote.c
gichan-jang: This is my mistake. I reverted this change. Thanks.

c/src/ml-api-service-remote.c
gichan-jang: As you mentioned, it is not possible to send a reply in the current pub/sub structure.
As suggested, the node type was changed to query server/client, so server can send the reply now.
I also added related tests, so please review them again.
Thanks for the suggestion.

c/src/ml-api-service-offloading.c
```diff
@@ -447,24 +453,44 @@ _mlrs_process_remote_service (nns_edge_data_h data_h, void *user_data)
...
-      remote_s->event_cb (event_type, NULL, remote_s->user_data);
+  if (mls && event_type != ML_SERVICE_EVENT_UNKNOWN) {
+    if (mls->cb_info.cb) {
+      mls->cb_info.cb (event_type, info_h, mls->cb_info.pdata);
     }
   }
 
+done:
+  if (info_h) {
+    ret = ml_information_destroy (info_h);
```
jaeyun-jung, line 490: unnecessary checking ret code here, call ml_information_destroy (info_h); and return previous error code.

c/src/ml-api-service-offloading.c
```diff
@@ -575,129 +605,156 @@ ml_service_remote_release_internal (ml_service_s * mls)
...
+  if (offloading_s == NULL) {
     _ml_error_report_return (ML_ERROR_OUT_OF_MEMORY,
         "Failed to allocate memory for the service handle's private data. Out of memory?");
   }
 
-  _mlrs_get_edge_info (option, &edge_info);
+  offloading_s->table =
+      g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  if (!offloading_s->table) {
+    _ml_error_report
```
jaeyun-jung, line 670: need return error code ML_ERROR_OUT_OF_MEMORY

c/src/ml-api-service-offloading.c
```diff
@@ -715,45 +772,75 @@ ml_service_remote_register (ml_service_h handle, ml_option_h option, void *data,
...
+  activate = json_object_get_string_member (service_obj, "activate");
+  if (activate) {
+    ret = nns_edge_data_set_info (data_h, "activate", activate);
+    if (NNS_EDGE_ERROR_NONE != ret) {
+      _ml_logi ("Failed to set activate in edge data.");
+    }
+  }
+  _in = (ml_tensors_data_s *) input;
+  ret =
+      nns_edge_data_add (data_h, _in->tensors[0].data, _in->tensors[0].size,
```
jaeyun-jung, line 801: Q: is it ok to append 1st memory? let's consider adding for loop (The number of tensor data may be larger than 1).

c/src/ml-api-service.c
```diff
@@ -328,10 +458,8 @@ ml_service_new (const char *config, ml_service_h * handle)
 
   object = json_node_get_object (root);
 
-  if (json_object_has_member (object, "single") ||
-      json_object_has_member (object, "pipeline")) {
-    service_type = ML_SERVICE_TYPE_EXTENSION;
-  } else {
+  service_type = _ml_service_get_type (object);
+  if (ML_SERVICE_TYPE_UNKNOWN == service_type) {
```
jaeyun-jung, line 462: duplicated error message (line 398 and 463)

songgot (approved):
LGTM

c/src/ml-api-service-offloading.c
gichan-jang: Thanks, Updated!

c/src/ml-api-service.c
gichan-jang: Thanks, Updated!

jaeyun-jung (approved):

PR #461, opened 2024-03-04
[Util] public function for tensor-data

jaeyun-jung:
Add new util functions,
1. clone tensor-data handle.
2. get tensors-info of tensor data frame.

This requires ACR for public release.


myungjoo (approved):

anyj0527 (approved):
LGTM👍 We needed this, cool

PR #463, opened 2024-03-04
[Agent] invalid handle type

jaeyun-jung:
Temporal bug fix before removing ml-agent from api repo.
Fix invalid handle type and check return ptr for gerror.


songgot (approved):
LGTM

wooksong (approved):

PR #473, opened 2024-03-25
[C-Api] use util function to copy tensor info

jaeyun-jung:
Code clean, use internal util function to create and copy tensor info handle.


gichan-jang (approved):
LGTM

PR #481, opened 2024-04-01
[action] replace TAOS-CI script for android build in API Level 34 @open sesame 04/04 20:31

niley7464:
This patch replaces TAOS-CI script for android build.

This PR requires https://github.com/nnstreamer/nnstreamer-android-resource/pull/41

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,67 @@
+name: Android Build Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        abi: ['armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64']
+    steps:
+    - uses: nttld/setup-ndk@v1
+      id: setup-ndk
+      with:
+        ndk-version: r25c
+        link-to-sdk: true
+    - name: Get gstreamer-android
+      run: |
+        mkdir -p ~/android/gst_root_android/
+        pushd ~/android/gst_root_android/
+        wget https://gstreamer.freedesktop.org/data/pkg/android/1.24.0/gstreamer-1.0-android-universal-1.24.0.tar.xz
+        tar -xf gstreamer-1.0-android-universal-1.24.0.tar.xz
+        popd
+    - name: Get nnstreamer
+      uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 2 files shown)


java/android/nnstreamer/build.gradle
```diff
@@ -1,16 +1,17 @@
...
 
 android {
-    compileSdkVersion 28
-    buildToolsVersion '28.0.3'
+    namespace "org.nnsuite.nnstreamer"
+    compileSdk 34
     defaultConfig {
         minSdkVersion 24
-        targetSdkVersion 28
+        targetSdkVersion 34
```
again4you, line 8: Question)
The reason why `targetSdkVersion` is 34 is that it is minimal required version of Google Play Store. right?

.github/workflows/android.yml
```diff
@@ -0,0 +1,67 @@
...
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        abi: ['armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64']
+    steps:
+    - uses: nttld/setup-ndk@v1
```
again4you, line 16: Question)
The latest version of setup-ndk is v1.4.2 and the version you uses is v1.0.0, which is released in 2021. Is there any reason to use v1?

again4you (commented):
I just added some questions. Except that, it looks good to me.

.github/workflows/android.yml
niley7464: https://github.com/nnstreamer/nnstreamer/blob/b251e9bc6ba87b66056d4bf2adc852038cb7e38e/.github/workflows/android.yml#L15

I used that version to use the same version as nnstreamer.

java/android/nnstreamer/build.gradle
niley7464: Yes we should use 34 version for Play Store.

wooksong (approved):
LG(reat)TM!

jaeyun-jung (approved):

jaeyun-jung:
@gichan-jang please update taos-ci on aws to avoid android lib build error.

gichan-jang:
I removed ml-api-android build from TAOS-CI.
But checklist is not properly updated yet.
Please do force push again. (If it's not refresh the TAOS-CI checklist, let's ignore it)

jaeyun-jung:
Now we removed android build on taos, merge this PR.

PR #482, opened 2024-04-02
[Tizen] Update unittest build script

gichan-jang:
Fix tizen auto coverage measurement failure.
Use default root path instead of environment variable. The model path of the config file is hard coded, so the wrong model path is passed to ml-service.

config_single_add.conf
```
{
    "single" :
    {
        "framework" : "tensorflow-lite",
        "model" : ["../tests/test_models/models/add.tflite"]
    }
}

```

### DIFF
--- a/packaging/run-unittest.sh
+++ b/packaging/run-unittest.sh
@@ -7,8 +7,7 @@
 ##
 setup() {
     echo "setup start"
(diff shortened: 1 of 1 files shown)


anyj0527 (approved):
LGTM✔

jaeyun-jung (approved):

PR #489, opened 2024-04-05
[android] do not remove android-resource dir

niley7464:
nnstreamer-android-resource directory is changed not to download using scripts.
So it shouldn't be removed.


jaeyun-jung (approved):

jaeyun-jung:
This PR does not affect tizen build, merge it now.

PR #492, opened 2024-04-05
[C-Api/Service] change ml-agent param

jaeyun-jung:
To remove unnecessary dbus-sig, add force-delete flag in ml-agent interface.


again4you (approved):
This PR caused a CI break issue because of the DBus interface change.
However, it will be applied with the patch of MLOps Agent so I will merge it.

PR #494, opened 2024-04-09
[Service] fix svace issues

jaeyun-jung:
Code clean, update function name for offloading and fix svace issues.


gichan-jang (approved):
LGTM

PR #496, opened 2024-04-11
[Test] remove unnecessary code

jaeyun-jung:
Code clean,
1. remove unnecessary header and precondition for test.
2. use prefix path for bin directory.
3. fix invalid thread callback function.


anyj0527 (approved):
LGTM👍

PR #500, opened 2024-04-18
[test] Let test daemon launches per testsuite not single testcase

anyj0527:
- Let the test daemon launches per testsuite. Launching it for each testcase is not practical.

### DIFF
--- a/tests/capi/unittest_capi_service_offloading.cc
+++ b/tests/capi/unittest_capi_service_offloading.cc
@@ -53,7 +53,6 @@ class MLOffloadingService : public ::testing::Test
 {
   protected:
   static GTestDBus *dbus;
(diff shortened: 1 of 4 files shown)


tests/capi/unittest_capi_service_offloading.cc
```diff
@@ -52,17 +52,16 @@ _get_config_path (const gchar *config_name)
 class MLOffloadingService : public ::testing::Test
 {
   protected:
-  GTestDBus *dbus;
+  static GTestDBus *dbus;
   int status;
```
jaeyun-jung, line 56: can you update this PR to delete this memver val (int status)? :)

jaeyun-jung (approved):

PR #514, opened 2024-05-02
[CodeClean] check null before using handle

jaeyun-jung:
Code clean, check null pointer of train-offloading handle before using it.


anyj0527 (approved):
LGTM

PR #519, opened 2024-05-09
[ml-service] Get string member from json object

gichan-jang:
Changed to check the existence of the jsom member before getting the value.
This change reduces unnecessary error messages that occur when setting optional ml information.


jaeyun-jung (approved):

niley7464 (approved):
LGTM 👍

PR #523, opened 2024-05-20
[Test] use available port number

jaeyun-jung:
Add code to set available port number when running offload test.

### DIFF
--- a/tests/capi/unittest_util.c
+++ b/tests/capi/unittest_util.c
@@ -32,7 +32,8 @@ int dlog_print (int level, const char *tag, const char *fmt, ...)
 /**
  * @brief Get available port number.
  */
-guint get_available_port (void)
+guint
+get_available_port (void)
 {
   struct sockaddr_in sin;
   guint port = 0;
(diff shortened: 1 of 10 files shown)


tests/capi/unittest_util.c
```diff
@@ -57,6 +58,52 @@ guint get_available_port (void)
...
+  }
+
+  g_close (status, NULL);
+
+  if (!g_file_get_contents (conf_base, &contents, NULL, NULL) ||
+      !g_file_set_contents (tmp_conf, contents, -1, NULL)) {
+    goto error;
+  }
+
+  cmd = g_strdup_printf ("sed -i '/__AVAIL_PORT__/c\\\"%s\" : \"%u\",' %s",
```
songgot, line 90: How about ?
sed 's/__AVAIL_PORT/port_number/g' conf_base

jaeyun-jung: updated, please review the changes.

songgot (approved):
LGTM

tests/capi/unittest_util.c
```diff
@@ -57,6 +58,73 @@ guint get_available_port (void)
...
+  }
+
+  g_close (status, NULL);
+
+  if (!g_file_get_contents (conf_base, &contents, NULL, NULL) ||
+      !g_file_set_contents (tmp_conf, contents, -1, NULL)) {
+    goto error;
+  }
+
+  cmd = g_strdup_printf ("sed -i 's/__AVAIL_PORT__/%u/g' %s", port, tmp_conf);
```
gichan-jang, line 110: We're using a special character @ that changes later in config, How about changing it from '__' to '@'?

songgot: In fact, since the same port as the remote device must be used, it seems that using each available port is only for TC.

jaeyun-jung: updated

PR #524, opened 2024-05-22
[Service] util function to request raw data

jaeyun-jung:
Add util function to request offloading service - send raw data.

### DIFF
--- a/c/src/ml-api-service-offloading.c
+++ b/c/src/ml-api-service-offloading.c
@@ -1055,7 +1055,8 @@ ml_service_offloading_stop (ml_service_h handle)
 }
 
 /**
- * @brief Register new information, such as neural network models or pipeline descriptions, on a offloading server.
+ * @brief Internal function to request service to ml-service offloading.
+ * Register new information, such as neural network models or pipeline descriptions, on a offloading server.
  */
 int
 ml_service_offloading_request (ml_service_h handle, const char *key,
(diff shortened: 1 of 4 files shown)


c/src/ml-api-service-offloading.c
```diff
@@ -1184,3 +1184,20 @@ ml_service_offloading_request (ml_service_h handle, const char *key,
     nns_edge_data_destroy (data_h);
   return ret;
 }
+
+/**
+ * @brief Register new information, such as neural network models or pipeline descriptions, on a offloading server.
```
songgot, line 1189: The contents of the brief are different from ml-api-service-offloading.h line 103.
Did you intend to?

c/src/ml-api-service-offloading.c
```diff
@@ -1184,3 +1184,20 @@ ml_service_offloading_request (ml_service_h handle, const char *key,
...
+int
+ml_service_offloading_request_raw (ml_service_h handle, const char *key,
+    void *data, size_t len)
+{
+  ml_tensors_data_s input;
+
+  /* Set internal data structure to send edge data. */
+  input.num_tensors = 1;
+  input.tensors[0].data = data;
+  input.tensors[0].size = len;
```
songgot, line 1200: Is there no need to set dimension and type?

jaeyun-jung: nns-edge does not check tensor shape or type. It handles data only.

c/src/ml-api-service-offloading.c
jaeyun-jung: I added comment, same contents w/ml_service_offloading_request() function.

c/src/ml-api-service-offloading.c
songgot: Thank you.

songgot (approved):
LGTM

PR #533, opened 2024-06-13
[ml-service] Fix bug in unittest_capi_service_training_offloading

songgot:
- The pipeline for training offloading is designed to perform training and validation, but nntrainer creates a model after finishing training, therefore validation time is necessary.

### DIFF
--- a/tests/capi/unittest_capi_service_training_offloading.cc
+++ b/tests/capi/unittest_capi_service_training_offloading.cc
@@ -200,10 +200,18 @@ TEST_F (MLServiceTrainingOffloading, trainingOffloading_p)
       "sender_start_thread", (GThreadFunc) sender_start_thread, sender_h);
   receive_thread = g_thread_new (
       "receiver_start_thread", (GThreadFunc) receiver_start_thread, receiver_h);
+
+  /**
+   * The pipeline is designed to perform training and validation.
+   * When nntrainer finishes training, the model is created.
+   * Therefore, validation time is given for TC.
+   */
   int loop = 120;
   while (loop--) {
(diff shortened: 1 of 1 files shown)


tests/capi/unittest_capi_service_training_offloading.cc
```diff
@@ -200,10 +200,18 @@ TEST_F (MLServiceTrainingOffloading, trainingOffloading_p)
...
+  /**
+   * The pipeline is designed to perform training and validation.
+   * When nntrainer finishes training, the model is created.
+   * Therefore, validation time is given for TC.
+   */
   int loop = 120;
   while (loop--) {
-    if (g_file_test (trained_model_path, G_FILE_TEST_EXISTS))
+    if (g_file_test (trained_model_path, G_FILE_TEST_EXISTS)) {
+      g_usleep (1000000);
```
niley7464, line 212: Is there any special reason why the time out is 1000000?
Is it a heuristic value?

myungjoo: It is a test case, not an actual applicatin or framework code. So... a heuristic value is fine.

Anyway, I recommend to let it print out something (reference: https://stackoverflow.com/questions/16491675/how-to-send-custom-message-in-google-c-testing-framework ) by
```
#define GTEST_COUT std::cerr << "[          ] [ INFO ]"
...
GTEST_COUT << "Waiting for training epoch completion. Timeout in " << loop * 1 << " seconds." << std::endl;
```

myungjoo (approved):

jaeyun-jung (approved):

PR #538, opened 2024-06-18
[Service] error code when parsing json

jaeyun-jung:
Change error code when parsing json string of config file.

### DIFF
--- a/c/src/ml-api-service.c
+++ b/c/src/ml-api-service.c
@@ -339,8 +339,8 @@ ml_service_new (const char *config, ml_service_h * handle)
   }
 
   if (!json_parser_load_from_data (parser, json_string, -1, &err)) {
-    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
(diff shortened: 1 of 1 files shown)


niley7464 (approved):
LGTM!

songgot (approved):
LGTM

c/src/ml-api-service.c
```diff
@@ -339,8 +339,8 @@ ml_service_new (const char *config, ml_service_h * handle)
   }
 
   if (!json_parser_load_from_data (parser, json_string, -1, &err)) {
-    _ml_error_report_return (ML_ERROR_INVALID_PARAMETER,
-        "Failed to parse configuration file. Parse error: %s",
+    _ml_error_report_return (ML_ERROR_IO_ERROR,
```
anyj0527, line 342: How about keep old lines? I think Json parse error means it succeed to load the json but fail to parse it. The existence of json file is checked already.

jaeyun-jung: agreed, close this pr.

PR #539, opened 2024-06-18
[android] Support tflite v2.16.1

anyj0527:
- Add tflite v2.16.1 support for android
- Deprecates old versions and leaves only v2.8.1 and v2.16.1
- In github actions, Enable tflite for x86_64 (for emulators) / Disable tflite for armeabi-v7a

See also https://github.com/nnstreamer/nnstreamer-android-resource/pull/44


niley7464:
I guess gradle couldn't find the flatbuffer library because I didn't update the mk file last time I built latest tensorflow with Bazel.
nnstreamer-android needs this PR to use tflite 2.16.1
https://github.com/nnstreamer/nnstreamer-android/pull/40

myungjoo (approved):

PR #543, opened 2024-06-25
JNI/Android: Specify an ABI that SNPE supports

wooksong:
Since SNPEv2 supports Android only for aarch64, this patch specifies the ABI for it in the Android.mk file.


songgot (approved):
LGTM

myungjoo (approved):

PR #546, opened 2024-07-11
[service] Allow no input / no output node in ext/conf

anyj0527:
- There are valid pipelines consists of no appsrc/tensor_sink.
- Let `ml_service_new` allow those cases.


myungjoo (approved):

myungjoo:
"invoke" calls of ML Service API are designed to have input and output.
You will need to document (e.g., at the MLAPI doxygen of "ml-api-service.h") how to handle such pipelines with ML Service API invoke(query) calls. (and how the invoke implementation will react)

jaeyun-jung (approved):

PR #551, opened 2024-07-22
 [CodeClean] leak case when parsing json str

jaeyun-jung:
Code clean, fix mem leak case when parsing json from string.


again4you (approved):
Nice Catch!

PR #552, opened 2024-08-21
[Refactor] Prepend `_` to  internal API name

gichan-jang:
To distinguish between internal and public APIs, prepend `_` to the internal APIs.


niley7464 (approved):
LGTM!
Thank you :)

jaeyun-jung (approved):

PR #562, opened 2024-09-09
[Common] refactor info list

jaeyun-jung:
Refactor functions for info-list, use linked list and add new function to append info.

### DIFF
--- a/c/src/ml-api-service-agent-client.c
+++ b/c/src/ml-api-service-agent-client.c
@@ -138,7 +138,7 @@ _build_ml_info_from_json_cstr (const gchar * jcstring, void **handle)
   }
 
   if (array)
-    ret = _ml_information_list_create (n, &_info_list);
+    ret = _ml_information_list_create (&_info_list);
   else
     ret = _ml_information_create (&_info);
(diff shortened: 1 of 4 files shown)


c/src/ml-api-service-agent-client.c
```diff
@@ -176,7 +183,7 @@ _build_ml_info_from_json_cstr (const gchar * jcstring, void **handle)
   } else {
     if (_info_list)
       ml_information_list_destroy (_info_list);
-    if (_info)
+    else if (_info)
       ml_information_destroy (_info);
```
songgot, line 187: in case of line 174, ```_info``` and ```_info_list``` need to free, right?

anyj0527 (approved):
LGTM👍

c/src/ml-api-service-agent-client.c
jaeyun-jung: In the changes, _info is added into _info_list. So If given json is array, destroying _info_list internally releases _info.

songgot: Thank very much :pray:

PR #569, opened 2024-10-11
[ml-service] Add staging header to prevent build error

gichan-jang:
Staging header will be added to Tizen 7.0.
Add empty staging header file to prevent build errors for users using the header file.


jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #575, opened 2024-10-31
[android] Add octet-stream decoder for android

niley7464:
This patch adds tensor-decder octet-stream mode for android.


anyj0527 (approved):
LGTM👍

niley7464:
https://github.com/nnstreamer/api/blob/f5b00af7f5102b5412bbe3489b7f8b91e5079df8/java/android/nnstreamer/src/main/java/org/nnsuite/nnstreamer/Pipeline.java#L59

The callback is designed to receive tensordata, so I don't think that mode is needed.

PR #577, opened 2024-11-07
[gcov] Fix invalid gcda files

gichan-jang:
Previously, a null file was created to cover cases where tests were not performed.
From the latest gcov version 14, this is treated as an invalid gcda file, resulting in an error.
Change to merge the initialized info and test result info.


anyj0527 (approved):
LGTM👍

jaeyun-jung (approved):

PR #582, opened 2024-11-20
[C-Api] fix mem leak case

songgot:
Code clean, fix mem leak case.


again4you:
* Origin: https://github.com/nnstreamer/api/commit/41bedd28b7576306af1c69817e35fe462de61705

Referring to the above commit, `_training_offloading_send_trained_model()` function part is removed. Please check the code whether the change is necessary or not.

PR #601, opened 2025-01-20
[Android] database for ml-agent

jaeyun-jung:
Add sqlite prebuilt library and set path to call ml-agent interface.


songgot (approved):
LGTM

java/android/nnstreamer/src/main/jni/nnstreamer-native-common.c
```diff
@@ -427,6 +441,28 @@ nnstreamer_native_initialize (JNIEnv * env, jobject context)
...
+{
+  _ml_logi ("Called native finalize.");
+
+  G_LOCK (nns_native_lock);
+
+#if defined(__ANDROID__)
+#if defined(ENABLE_ML_AGENT)
+  ml_agent_finalize ();
+#endif
+#endif
```
songgot, line 458: It would be nice to modify it as follows. What do you think?

```
#if defined(__ANDROID__) && defined(ENABLE_ML_AGENT)
  ml_agent_finalize ();
#endif
```

jaeyun-jung: For future work, case if android and not ml-agent.

gichan-jang (approved):
LGTM

PR #607, opened 2025-02-05
[api] Let single handle flexible filter

anyj0527:
- Let single API handle flexible filter.

### DIFF
--- a/c/src/ml-api-inference-single.c
+++ b/c/src/ml-api-inference-single.c
@@ -137,6 +137,7 @@ typedef struct
   gboolean invoking;                  /**< invoke running flag */
   ml_tensors_data_h in_tensors;       /**< input tensor wrapper for processing */
   ml_tensors_data_h out_tensors;      /**< output tensor wrapper for processing */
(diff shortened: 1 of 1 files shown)


jaeyun-jung (approved):

c/src/ml-api-inference-single.c
```diff
@@ -1083,6 +1091,21 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
     goto error;
   }
 
+  /* handle flexible single */
+  /**
+   * Set invoke_dynamic as TRUE if the given nnfw do invoke_dynamic
+   *
+   * if (info->nnfw == ML_NNFW_TYPE_EXECUTORCH_LLAMA || info->nnfw == ML_NNFW_TYPE_LLAMACPP) {
+   *   invoke_dynamic = TRUE;
```
myungjoo, line 1099: It should be set by the subplugin or pipeline user, not by the API implementation.
The API implementation (and nnstreamer core) should NOT aware such details.

songgot (approved):
LGTM

PR #617, opened 2025-03-06
[Android] common path for external libs

jaeyun-jung:
Set common path for external libraries on android build.
This PR will build android library including same external files - SNPE, QNN, and TF-lite QNN delegate.


myungjoo (approved):

again4you (approved):
LGTM!

PR #623, opened 2025-03-25
[Android] gst version

jaeyun-jung:
It is hard to fix error with old version of gstreamer and build tool.
Set min version of gstreamer 1.24 and ndk r25c to build native library.


myungjoo:
```
It is hard to fix error with old version of gstreamer and build tool.
Set min version of gstreamer 1.24 and ndk r25c to build native library.
```
Where is it? (setting min version, returning error if not satisfied)

anyj0527 (approved):
LGTM✔

jaeyun-jung:
@myungjoo 
I pushed the changes to check gstreamer version when building android library.

myungjoo (approved):

PR #636, opened 2025-06-23
[ml service] Support asynchronous output from sub-plugin in ML service extension single API

songgot:
- Add invoke async callback to receive results from sub-plugins asynchronously.


myungjoo:
llama is way too big for a unit test. can you add a small dummy model to run a unit test? (e.g., <1MB). There are other unit test cases that use such small dummy models.

songgot:
> llama is way too big for a unit test. can you add a small dummy model to run a unit test? (e.g., <1MB). There are other unit test cases that use such small dummy models.
> 
> We execute all unit test cases for EVERY git pull-request, Tizen build / submit-request, ...

Okay, I will add next PR. Thank you.

c/src/ml-api-inference-single.c
```diff
@@ -1088,6 +1088,20 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
     g_object_set (filter_obj, "custom", info->custom_option, NULL);
   }
 
+  if (single_h->klass && info->invoke_async) {
+    if (info->invoke_async_cb != NULL && info->invoke_async_data!= NULL) {
+      NNSFilterInvokeAsyncCallback invoke_async_cb =
+          (NNSFilterInvokeAsyncCallback) info->invoke_async_cb;
+      single_h->klass->set_invoke_async_callback (single_h->filter,
```
jaeyun-jung, line 1095: We need update of nnsteamer into tizen first. After update, it is hard to fix both nnstreame and api repo.
So, let's focus the function and callback of async-invoke in nnstreamer repo now.

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -496,6 +496,31 @@ TEST_REQUIRE_TFLITE (MLServiceExtension, scenarioConfigLlamacpp)
   EXPECT_EQ (status, ML_ERROR_NONE);
 }
 
+/**
+ * @brief Usage of ml-service extension API.
+ */
+TEST_REQUIRE_TFLITE (MLServiceExtension, scenarioConfigLlamacppAsync)
```
jaeyun-jung, line 502: Invalid prefix TEST_REQUIRE_TFLITE
This testcase does not require tf-lite.

c/src/ml-api-inference-single.c
```diff
@@ -1078,14 +1075,33 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
...
   g_free (hw_name);
 
+  if (info->invoke_dynamic)
+    single_h->format = _NNS_TENSOR_FORMAT_FLEXIBLE;
+
   if (info->custom_option) {
     g_object_set (filter_obj, "custom", info->custom_option, NULL);
   }
 
+  if (single_h->klass && info->invoke_async) {
```
jaeyun-jung, line 1091: single_h->klass always is not null here.

PR #638, opened 2025-07-16
[ml service] Add Flare as a new nnfw type

songgot:
- Implemented support for 'flare' nnfw type in ML service API
- Included test cases to validate flare functionality


myungjoo:
Remove unrelated commits. Make it possible to be reviewed and merged independently from other topics

c/src/ml-api-inference-single.c
```diff
@@ -979,16 +980,25 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
...
-  status = _ml_validate_model_file ((const char **) list_models, num_models,
-      &nnfw);
-  if (status != ML_ERROR_NONE) {
-    _ml_error_report_continue
-        ("Cannot validate the model (1st model: %s. # models: %d). Error code: %d",
-        list_models[0], num_models, status);
-    g_strfreev (list_models);
-    return status;
+  /**
+   * Note : OpenVINO and flare use the bin extension.
```
jaeyun-jung, line 984: well, I suggest updating _ml_validate_model_file() function. plz do not update this block.
We simply check file extension and pass the validation.
```
int _ml_validate_model_file()
{
...
  /** @todo Make sure num_models is correct for each nnfw type */
  switch (*nnfw) {
...
    case ML_NNFW_TYPE_FLARE:
      if (!g_str_equal (file_ext[0], ".bin")) {
        status = ML_ERROR_INVALID_PARAMETER;
      }
      break;
}
```

songgot: Currently ML_NNFW_TYPE_FLARE is not included ml_nnfw_type_e.
It is defined to avoid ACR.
#define ML_NNFW_TYPE_FLARE 23
when code is changed

```
../c/src/ml-api-inference-single.c: In function ‘_ml_validate_model_file’:
../c/src/ml-api-inference-single.c:2051:5: error: case value ‘23’ not in enumerated type ‘ml_nnfw_type_e’ [-Werror=switch]
 2051 |     case ML_NNFW_TYPE_FLARE:
      |     ^~~~
```

songgot: So I will change the code as below.
```
  /** @todo Add checking ML_NNFW_TYPE_FLARE to below switch */
  if (*nnfw == ML_NNFW_TYPE_FLARE) {
    if (!g_str_equal (file_ext[0], ".bin")){
      _ml_error_report("Flare accepts .bin file only. Please support correct file extension. You have specified: \"%s\"", file_ext[0]);
      status = ML_ERROR_INVALID_PARAMETER;
    }
    goto done;
  }
```

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -486,7 +496,33 @@ TEST (MLServiceExtension, scenarioConfigLlamacpp)
   status = ml_service_new (config, &handle);
   ASSERT_EQ (status, ML_ERROR_NONE);
 
-  _extension_test_llamacpp (handle, FALSE);
+  _extension_test_llm (handle, FALSE, NULL, 5000000U);
```
hj210, line 499: How about use 'sleep_us_5xx(?)' instead of '5000000U'

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -486,7 +496,33 @@ TEST (MLServiceExtension, scenarioConfigLlamacpp)
...
+    g_critical ("Skipping scenarioConfigFlare test due to missing model file.Please download model file");
+    return;
+  }
+
+  g_autofree gchar *config = get_config_path ("config_single_flare.conf");
+
+  status = ml_service_new (config, &handle);
+  ASSERT_EQ (status, ML_ERROR_NONE);
+
+  _extension_test_llm (handle, FALSE, input_file, 40000000U);
```
hj210, line 525: How about use 'sleep_us_4xx(?)' instead of '4000000U'

tests/capi/unittest_capi_service_extension.cc
songgot: Okay, good idea.

c/include/nnstreamer-tizen-internal.h
```diff
@@ -19,6 +19,8 @@
 extern "C" {
 #endif /* __cplusplus */
 
+#define ML_NNFW_TYPE_FLARE 23 /**< FLARE framework */
```
myungjoo, line 22: You should write a note in the comment describing why this is defined here, not in ```ml-api-common.h```, and when/under-which-codition this  will be migrated to there. Otherwise, months later, we won't know why this is here and we won't move it to the proper position.

Plus, as long as Tizen-next is already submited, can you move it to the proper file and prepare ACR for 2026-M1?

c/src/ml-api-inference-single.c
```diff
@@ -2073,7 +2080,21 @@ _ml_validate_model_file (const char *const *model,
     file_ext[i] = g_ascii_strdown (pos, -1);
   }
 
-  /** @todo Make sure num_models is correct for each nnfw type */
+  /**
+   * @todo Currently ML_NNFW_TYPE_FLARE is defined temporarily to avoid ACR.
+   * Move checking ML_NNFW_TYPE_FLARE to below switch statement.
+   */
+  if (*nnfw == ML_NNFW_TYPE_FLARE) {
+    if (!g_str_equal (file_ext[0], ".bin")) {
```
hj210: How about separating the validation logic into a separate function according to Clean Code SRP, as shown below?
```cpp
    if (*nnfw == ML_NNFW_TYPE_FLARE) {
        return _validate_flare_model(model, num_models);
    }
```

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -406,12 +406,12 @@ _extension_test_imgclf (ml_service_h handle, gboolean is_pipeline)
...
   } while (0)
 
+
 /**
  * @brief Callback function for scenario test.
  */
 static void
-_extension_test_llamacpp_cb (
-    ml_service_event_e event, ml_information_h event_data, void *user_data)
+_extension_test_llm_cb (ml_service_event_e event, ml_information_h event_data, void *user_data)
```
hj210: It seems that the functions _extension_test_llamacpp_cb and _extension_test_llm_cb have the same structure.
They could likely be refactored into a common test function as shown below.
For example, branching based on framework_name.
```cpp
static inline void
_extension_test_llm_framework(const gchar *config, const gchar *input_file, 
                             guint sleep_us, gboolean is_pipeline, 
                             const gchar *framework_name)
{
    // common test function
}
```

hj210 (commented):
The commit purpose and changes are clearly described, and the test code appears to be well-written. Here are a few suggestions to improve maintainability and code quality:

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -494,26 +506,44 @@ _extension_test_llamacpp (const gchar *config, gboolean is_pipeline)
  */
 TEST (MLServiceExtension, scenarioConfigLlamacpp)
 {
-  skip_llamacpp_tc ("scenarioConfigLlamacpp");
+  skip_llm_tc ("scenarioConfigLlamacpp", "llama-2-7b-chat.Q2_K.gguf");
 
   g_autofree gchar *config = get_config_path ("config_single_llamacpp.conf");
 
-  _extension_test_llamacpp (config, FALSE);
+  _extension_test_llm (config, NULL, 5000000U, FALSE);
```
hj210: How about using constant for testing time to clarify it?
For example,
#define LLAMACPP_TEST_TIMEOUT_US 5000000U

c/src/ml-api-inference-single.c
songgot: Good idea. However, this code should be moved to the switch statement below after the ACR and managed together.

tests/capi/unittest_capi_service_extension.cc
songgot: Good idea, but let's keep it like other TCs.

tests/capi/unittest_capi_service_extension.cc
```diff
@@ -494,26 +506,44 @@ _extension_test_llamacpp (const gchar *config, gboolean is_pipeline)
...
+ */
+TEST (MLServiceExtension, scenarioConfigFlare)
+{
+  g_autofree gchar *input_file = g_strdup ("flare_input.txt");
+
+  skip_llm_tc ("scenarioConfigFlare", "sflare_if_4bit_3b.bin");
+
+  g_autofree gchar *config = get_config_path ("config_single_flare.conf");
+
+  _extension_test_llm (config, input_file, 40000000U, FALSE);
```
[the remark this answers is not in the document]
songgot: Good idea, but let's keep it like other TCs.

PR #641, opened 2025-08-04
[C-Api] flag to set invoke mode

jaeyun-jung:
Code clean, add flag to check invoke mode is dynamic or async.


c/src/ml-api-common.c
```diff
@@ -980,26 +983,23 @@ ml_tensors_data_create (const ml_tensors_info_h info, ml_tensors_data_h * data)
   }
 
   _info = (ml_tensors_info_s *) info;
-  if (_info->info.format == _NNS_TENSOR_FORMAT_FLEXIBLE) {
+  if (_info->info.format == _NNS_TENSOR_FORMAT_STATIC) {
+    for (i = 0; i < _data->num_tensors; i++) {
+      status = _ml_tensor_data_alloc (_data, i, _data->tensors[i].size);
```
songgot, line 988: I think it would be good if you could also modify the following.
Line 953
gint status = ML_ERROR_STREAMS_PIPE;  --> ML_ERROR_NONE;

songgot (approved):
LGTM

PR #645, opened 2025-08-21
[CodeClean/Test] invalid option

jaeyun-jung:
Code clean, add definition on ubuntu test and fix invalid property value.


gichan-jang (approved):
LGTM

PR #685, opened 2026-08-19
[C-Api] handle message callback

jaeyun-jung:
Handle message callback of ml-service.
- add callback to pass message from pipeline bus.
- send message when failed to invoke a model in ml-service thread.


gichan-jang (approved):
LGTM

PR #686, opened 2026-08-21
[C-Api] release allocated data

jaeyun-jung:
Release allocated memory when ml-tensors-data handle is destroyed.


gichan-jang (approved):
LGTM

PR #687, opened 2026-08-25
[C-Api] define new event for message

jaeyun-jung:
Define new enum - message in ml-service event.
This event is passed to user to notify an error or a change of ml-service.


gichan-jang (approved):
LGTM

PR #688, opened 2026-09-03
[CI] Restore the per-function Doxygen check

myungjoo:
Port of nnstreamer/nnstreamer#4914 (issue nnstreamer/nnstreamer#4908).

## What was wrong

`.github/workflows/static.check.scripts/doxygen-tag.sh` here is byte-identical to the nnstreamer copy: `local function_check_flag="f+p"` at script top level, which bash rejects, and no `set -e`, so `ctags -x --c-kinds=` ran with an empty kind list and the per-function `@brief` check has passed vacuously since the Actions import. `$report_path` was never set (an "ambiguous redirect" per file on stderr) and `$brief` was not reset between files.

## What this PR does

Takes the corrected checker and its self-test from nnstreamer/nnstreamer#4914 verbatim, and adds the self-test step to `static.check.yml` with the same merge-ref guard:

- definitions are checked everywhere, prototypes only in headers;
- the trailing `/**< ... */` form is recognised (declaration line, a line opening with `/**<` right after it, or after an inline body);
- block comments are tracked so a `*` at the start of a wrapped parameter line is not taken for a comment continuation;
- `report_path` defaults to `/dev/null`, `brief` resets per file;
- `test_doxygen_tag.sh` generates 18 fixtures into a temp dir; all 18 fail against the checker on `main` and pass here.

## Backlog

None. Under the corrected scope the tree is already clean (48 C/C++ files, zero findings with Exuberant Ctags 5.9, the CI version), so the check turns on without touching any source file.

### DIFF
--- a/.github/workflows/static.check.scripts/doxygen-tag.sh
+++ b/.github/workflows/static.check.scripts/doxygen-tag.sh
@@ -28,11 +28,41 @@ fi
 
 files=$1
 failed=0
+report_path=${report_path:-/dev/null}
 
 if [ ! -f $files ]; then
   echo "::error The file $files does not exists."
   exit 1
 fi
+
+##
+# @brief Tell whether the declaration starting at a line carries a trailing
+#        "/**<" Doxygen comment, on the declaration itself or right after it.
+# @param $1 1-based line number where ctags placed the function
+# @return 0 when such a comment is found, 1 otherwise
+# Reads the global array file_lines (0-based). A declaration ends at the first
+# ';' outside braces, or at the '}' that closes an inline body. The comment may
+# also open the next line; a "/**<" later on that line belongs to whatever is
+# declared there, not to this one.
+has_trailing_doc() {
+  local -i i=$1-1
+  local -i last=${#file_lines[@]}-1
+  local -i depth=0
+  local l opens closes
+  while [[ $i -le $last ]]; do
+    l=${file_lines[$i]}
+    [[ $l == *"/**<"* ]] && return 0
+    opens=${l//[^\{]/}
+    closes=${l//[^\}]/}
+    depth+=${#opens}-${#closes}
+    if [[ $depth -le 0 && ( $l == *";"* || $l == *"}"* ) ]]; then
+      [[ $i -lt $last && ${file_lines[$i+1]} =~ ^[[:space:]]*/\*\*\< ]] && return 0
+      return 1
+    fi
+    i+=1
+  done
+  return 1
+}
 
 echo "::group::Doxygen tag check started"
 
@@ -69,14 +99,20 @@ for file in `cat $files`; do
               # Checking tags for each function
               if [[ $advanced == 1 ]]; then
                   declare -i idx=0
+                  brief=0
+                  in_comment=0
                   function_positions="" # Line number of functions.
                   structure_positions="" # Line number of structure.
+                  mapfile -t file_lines < "$file"
 
-                  local function_check_flag="f+p" # check document for function and prototype of the function
-
-                  if [[ $pr_doxygen_check_skip_function_definition == 1 && $file != *.h ]]; then
-                      function_check_flag="p" # check document for only prototypes of the function for non-header file
-                  fi
+                  # Definitions are checked everywhere. Prototypes are checked
+                  # only in headers, where the declaration is the documented
+                  # form; a forward declaration in a .c/.cc is documented at
+                  # its definition.
+                  case $file in
+                      *.h|*.hh|*.hpp ) function_check_flag="f+p" ;;
+                      * ) function_check_flag="f" ;;
+                  esac
 
                   # Find line number of functions using ctags, and append them.
                   while IFS='' read -r line || [[ -n "$line" ]]; do
@@ -97,7 +133,7 @@ for file in `cat $files`; do
                       # Check if a function has @brief tag or not.
                       # To pass correct line number not sub number, keep space " $idx ".
                       # ex) want to pass 143 not 14, 43, 1, 3, 4
-                      if [[ $function_positions =~ " $idx " && $brief -eq 0 ]]; then
+                      if [[ $function_positions =~ " $idx " && $brief -eq 0 ]] && ! has_trailing_doc $idx; then
                           echo "[ERROR] File name: $file, $idx line, `echo $line | cut -d ' ' -f1` function needs @brief tag "
                           failed=1
                       fi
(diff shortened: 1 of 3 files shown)


myungjoo:
[Automated review by a separate Claude agent, relayed to this PR]

Scope: read nnstreamer/nnstreamer#4908 and the full review history of nnstreamer/nnstreamer#4914 (`gh pr view 4914 --repo nnstreamer/nnstreamer --comments`, 4 review rounds), diffed this PR (`upstream/main...HEAD`, 3 files, +325/-8) and byte-compared `doxygen-tag.sh`/`test_doxygen_tag.sh` against nnstreamer's final approved commit `4490a668` (identical modulo CRLF), read `static.check.yml`'s new step and the rest of the job, and independently re-verified the PR's "zero backlog" claim against this repo's own 48 C/C++ files (ctags is not available in my sandbox, so I wrote and sanity-checked a Python heuristic against the checker's actual state-machine rules, cross-checked with `gh pr checks` / `gh run view --log`). Also checked `nnstreamer/api`'s branch protection via the GitHub API.

## Does it solve the stated problem

Yes. `doxygen-tag.sh` here was byte-identical to nnstreamer's pre-#4914 copy, carrying the same top-level `local function_check_flag="f+p"` (invalid outside a function, silently emptying `ctags`'s kind list with no `set -e` to catch it), the same unset `$report_path`, and the same missing per-file reset of `$brief`. This PR takes nnstreamer's already-fixed, 4-round-reviewed version verbatim — I confirmed `diff -u --strip-trailing-cr` between this PR's two scripts and nnstreamer PR #4914's head (`4490a668`) is empty. That version's design (definitions checked everywhere, prototypes only in `.h`/`.hh`/`.hpp`; trailing `/**<` recognition on same line, next line, or after an inline body; real block-comment tracking via `in_comment` instead of a bare-`*` heuristic) was independently traced and stress-tested across 4 review rounds on the origin PR, including two real counterexamples found and fixed in-tree (`gsttensor_converter.c`'s wrapped-pointer prototype, `Looper::post` getting credited for a sibling's trailing `/**<`). Nothing in this port reopens either of those.

## Backlog = 0 claim — verified independently

I could not run `ctags` locally (not installed, no sudo in my sandbox), so instead of trusting the PR body I wrote a Python approximation of the checker's own rules (K&R return-type-on-own-line handling, `@brief`/`@copydoc` block detection, trailing `/**<` on the same/next line) and sanity-checked it against a synthetic documented/undocumented pair before trusting its output on the real tree.

- Whole-tree scan (48 files): 1141 candidate function definition/declaration sites, 0 flagged as undocumented.
- Header-prototype-only scan (`f+p` scope, 9 header files): 216 candidate prototypes, 0 flagged.
- Every one of the 438 GTest `TEST (suite, case)` macros across `tests/capi/*.cc` (these get tagged as ordinary functions by ctags' generic C parser) is immediately preceded by a comment block containing `@brief`, verified with a separate direct script, not just the heuristic above.
- Every named-tag `struct`/`class` in the tree (`_ml_pipeline`, `_ml_pipeline_element`, `_ml_pipeline_common_elem`, `_type_int`, `_type_string`, `_type_element`, `_conf_detail`, `_camera_conf`) has its own leading `@brief`.
- This matches the PR's own CI: the self-test step ran with the real `exuberant-ctags` the "Preparing step 2" installs, and all 18 `PASS:` lines plus `test_doxygen_tag.sh: all checks passed.` appear in the live log of run `33715005844`.

I did not find a counterexample. The claim holds as far as I can verify it.

## Self-test / CI wiring

`test_doxygen_tag.sh` (18 fixtures) is a solid regression suite: it pins the negative control (undocumented definition must fail — the exact property that was silently broken for years), both scope halves (`f` vs `f+p`), all three `/**<` placements plus the two adjacency traps found in the origin review (`trailing_belongs_to_next.hh`, `pointer_line.h`/`wrapped_pointer.h`), the per-file state reset in both directions, and asserts zero stderr output (which is how the original `local` bug and the unset `report_path` actually manifested). It generates fixtures into `mktemp -d` rather than committing them, so an intentionally-undocumented fixture never lands in a future PR's `$changed_file_list`.

**Positive note on the port, not an issue:** in nnstreamer's `static.check.yml` this same self-test step's comment says "needs the same merge-ref guard as the self-test steps below" — there it's one of five sibling self-test steps sharing the pattern. `nnstreamer/api`'s `static.check.yml` has no other self-test step, and the author rewrote the comment here to be self-contained (explaining *why* the merge-ref-vs-checkout-ref mismatch can happen, rather than pointing at nonexistent siblings). I checked — there is genuinely no other `diff-filter=D`/self-test step anywhere in this repo's `.github/workflows/`, so the rewritten comment is the correct call, not an oversight.

## should-fix (not blocking this PR — outside its diff, but answers a question this review was asked to check)

**`nnstreamer/api`'s branch protection on `main` does not include "Static checks" as a required status check** (`gh api repos/nnstreamer/api/branches/main/protection` → `required_status_checks.contexts: ["DCO"]` only). This predates this PR — the "Static checks" job already existed with a dozen other checkers before this change — and it isn't something this PR's diff can fix (the repo has no `.github/settings.yml` or other repo-as-code branch-protection file; it's a GitHub UI/API-only setting). But it directly bears on the review question "can CI failing actually block a future regression from merging": today, a red `Static checks` job (including a future regression this exact self-test would catch) does not by itself block the merge button, only a red/missing `DCO` does. Recommend the maintainer add `Static checks` to the required status checks for `main` in a follow-up (not part of this PR).

## Informational / no PR action needed

- **`Spell Check with Typos` is currently failing on this PR's CI**, but on a pre-existing typo ("gievn" → "given") in `c/src/ml-api-service-offloading.c:306`, introduced in commit `751f51b6` (2023-06-30), a file this PR does not touch. Confirmed via `git blame` against `upstream/main`. Unrelated to this change.
- The known, already-accepted-upstream heuristic limits of `has_trailing_doc`/`in_comment` (can't distinguish a real `/*` from one inside a string literal or a `x / *p` expression without a space) carry over unchanged from the already-reviewed nnstreamer version; no live instance of either pattern exists in this tree today (checked by grep), same conclusion the origin reviews reached for nnstreamer's tree.
- `has_trailing_doc` accepts any `/**<` without requiring `@brief`/`@copydoc` inside it, which is a slightly lower bar than the leading-comment form — again unchanged from the reviewed-and-approved upstream version, and consistent with how this repo's own headers already use `/**<` on enum/struct members.

## Size / scope

Proportionate and minimal: 3 files, all under `.github/workflows/`, no source file touched (the tree already being clean under the corrected scope means there's no comment-only backlog-clearing commit needed here, unlike the origin PR's 34 files). No unrelated module is touched. Single commit, DCO-signed, subject line follows `[Component] Summary` convention, body explains what/why and credits the upstream fix.

## Architecture / API docs

None needed. CI-script-only change; this repo has no `CLAUDE.md`/`AGENTS.md`/`Documentation/` referencing the doxygen-check policy that this PR's scope choices would need to update.

## CI status at review time

`Static checks`: pass (run `33715005844`), self-test 18/18 `PASS`, doxygen-tag check itself ran clean over the 2 changed files. `DCO`: pass. `Spell Check with Typos`: fail, pre-existing/unrelated (see above). `build (armeabi-v7a)`: pass. Remaining `build (*)`/Tizen GBS jobs were still pending/queued at review time; this PR touches no source or build file, so I have no reason to expect them to be affected, but I did not wait for them to finish.

## Verdict

Mergeable on content. The restored checker and its self-test are a faithful, verified port of the version that survived 4 rounds of review upstream, the "zero backlog" claim checks out under independent verification (including the highest-volume risk category, the 438 GTest macros), and the PR is minimal and correctly scoped. The one should-fix (branch protection not actually requiring "Static checks") is real and worth a maintainer follow-up, but it's a repository setting outside this PR's diff, not a reason to hold this PR.

PR #692, opened 2026-09-05
[Debian/Build] Build ml-api on Ubuntu 26.04 (gcc fallback, C++17)

myungjoo:
## Why

`ppa:nnstreamer/ppa` now publishes nnstreamer, nnstreamer-edge, ssat, tensorflow2-lite and edgetpu for Ubuntu 26.04 (resolute), but `ml-api` cannot follow: the Launchpad recipe `nnstreamer-ml-api-daily` stops at the build-dependency stage on 24.10 and later because every compiler alternative in `debian/control` (`gcc-9 | gcc-8 | ... | gcc-5`) has been dropped from the archive. Once past that, all four gtest binaries fail to compile on 26.04 because GoogleTest 1.17 refuses anything older than C++17.

## What

- `debian/control`: append the unversioned `gcc` as the last compiler alternative, the same fallback nnstreamer (#4895) and nnstreamer-edge (nnstreamer/nnstreamer-edge#256) already carry. 22.04 and 24.04 still resolve `gcc-9` first, so nothing changes there.
- `meson.build`: `cpp_std=c++14` -> `c++17`. The Android build already passes `-std=c++17` and nnstreamer has been on C++17 for years.

## How it was verified

Full packaging build in an `ubuntu:26.04` container with `ppa:nnstreamer/ppa` enabled, `mk-build-deps` from the modified control, then `dpkg-buildpackage -us -uc -b` (gcc 15.2.0, meson 1.10, openjdk 11 from the archive). Without the second commit the build stops with

```
/usr/include/gtest/internal/gtest-port.h:273:2: error: #error C++ versions less than C++17 are not supported.
```

in `unittest_capi_inference`, `unittest_capi_inference_single`, `unittest_capi_inference_latency` and `unittest_capi_datatype_consistency`. With both commits the build completes, `override_dh_auto_test` (`packaging/run_unittests.sh`) runs, and all eight `.deb`s are produced.

No functional change, so no new test case; the existing unit tests are what the C++17 change makes buildable again on 26.04.

Part of the 26.04 PPA work tracked in nnstreamer/nnstreamer#4902. After this merges, the Launchpad recipe needs Resolute added and a build requested (I can do that).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: ml-api
 Section: libs
 Priority: optional
 Maintainer: MyungJoo Ham <myungjoo.ham@samsung.com>
-Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4),
+Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4) | gcc,
  ninja-build, meson (>=0.50), debhelper (>=9),
  libgstreamer1.0-dev, libgstreamer-plugins-base1.0-dev, gstreamer1.0-plugins-base,
  nnstreamer, nnstreamer-dev, nnstreamer-dev-internal, openjdk-11-jdk,
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project('ml-api', 'c', 'cpp',
     'werror=true',
     'warning_level=1',
     'c_std=gnu89',
-    'cpp_std=c++14'
+    'cpp_std=c++17'
   ]
 )
 


myungjoo:
[Automated review by a separate Claude agent, relayed to this PR by the author's account]

Reviewed both commits (7ffa297, 492e55e) against a fresh checkout of the branch.

**Verdict: approve.** No blocking findings; one minor commit-message nit.

### What was checked

- **C++ surface affected by the standard bump.** The only C++ sources in the repo are the nine gtest binaries under `tests/capi/*.cc`; every shipped library (`c/src`, `java/android/.../jni`) is C. So `cpp_std=c++17` changes nothing about installed artifacts or ABI. A grep for constructs removed in C++17 (`std::auto_ptr`, `random_shuffle`, dynamic exception specifications, `register`, `unary_function`/`bind1st`/`ptr_fun`) finds nothing in the test sources.
- **Other build files do not disagree.** `java/android/nnstreamer/src/main/jni/Application.mk` already sets `APP_CPPFLAGS := -std=c++17`; there is no CMake, and no `cpp_args`/`override_options` anywhere in the meson tree that would pin a different standard.
- **Tizen GBS.** `packaging/machine-learning-api.spec` strips `-std=gnu++11` from `CXXFLAGS` explicitly so that meson decides the standard, so the bump does propagate to Tizen builds. That is the same toolchain nnstreamer has been building with `cpp_std=c++17` for years, so no new risk there.
- **22.04 / 24.04.** Both series still carry `gcc-9` in universe, and the new `| gcc` is appended last, so apt resolution on jammy/noble is unchanged. gcc-9 with `-std=c++17 -Werror` is exercised today by nnstreamer's own jammy pdebuild, and this PR's jammy pdebuild job will confirm it for these test files.
- **meson floor.** `meson_version: '>=0.50.0'` accepts `c++17` as a `cpp_std` value, and meson has had the `c++17` -> `c++1z` fallback for older compilers since before 0.50.
- **Bisectability.** Commit 1 alone only lets 26.04 get further (to the gtest error); it cannot regress any series that built before. Commit 2 alone is harmless on every series. Either order is safe.
- **Test policy.** Build-script-only change; the PR body states why no new test case is added, as CLAUDE.md requires.
- **Commit hygiene.** Subjects follow `[Component] Summary` (`[Debian]`, `[Build]`); each commit has exactly one `Signed-off-by` for the author plus a `Co-Authored-By` trailer; no duplicate sign-off.

### Findings (ranked)

1. **Minor / commit message accuracy (492e55e).** The body says "gcc 7 and later, the oldest compiler the packaging accepts, support it", but `debian/control` still lists `gcc-6 | gcc-5 (>=5.4)` as alternatives. On those compilers meson would fall back to `-std=c++1z`, which gcc 5/6 only partially implement. In practice this is moot (no supported series ships gcc-5/6, and the gtest that needs C++17 is not on any series that has them), but the sentence is wrong as written. Suggest rewording to something like "every compiler a current series can resolve (gcc-9 and later) supports it", or, separately, pruning the gcc-5/6 alternatives if you want the claim to hold literally. Not worth blocking on.

2. **Informational, not a defect.** No CI job builds on 24.10+/26.04, so the 26.04 verification is manual-only and will stay that way until the Launchpad recipe gets Resolute added (which the PR body already plans). If you want it covered in-repo later, the pdebuild matrix in `.github/workflows/pdebuild.yml` would need a second `distroname` entry; out of scope for this PR.

Nothing else. The change is the minimum needed and mirrors what nnstreamer (#4895) and nnstreamer-edge (#256) already merged.

myungjoo:
[Relayed on behalf of the author by a Claude agent]

Thanks. Finding 1 is fixed: the second commit's message now says "every compiler a current series can resolve from the packaging, gcc-9 and later, supports it" instead of the wrong gcc-7 claim (commit message only, no content change; force-pushed to the same branch). The gcc-5/6 alternatives are left as they are in this PR since no series that still has them is affected.

Finding 2 is understood: 26.04 coverage comes from the Launchpad recipe once Resolute is added there, which is the next step after this merges.

myungjoo:
[Automated re-review by a separate Claude agent, relayed to this PR by the author's account]

Re-reviewed after the force-push that rewords commit 792c886's message (7ffa297 unchanged). Verified directly against `origin/fix/ubuntu-26.04-gcc` rebased on `upstream/main`, not just the GitHub diff view.

**Verdict: approve.** The prior minor finding is fixed, and no new issues were found on independent re-verification.

### Prior finding re-checked

- **Finding 1 (commit message accuracy) — confirmed fixed.** Commit 792c886 now reads "Every compiler a current series can resolve from the packaging, gcc-9 and later, supports it," replacing the earlier "gcc 7 and later" claim. This is accurate: apt resolves `gcc-9` first on 22.04/24.04, and falls through to the newly-added unversioned `gcc` (→ gcc-13/14/15) only on 24.10+/26.04 — every path lands on gcc-9 or newer. The stale `gcc-6 | gcc-5 (>=5.4)` alternatives are left untouched, as the author noted, since no currently supported series resolves them; this doesn't affect the truth of the reworded sentence. Message-only change, `oid` list confirms no content diff: `debian/control`/`meson.build` are byte-identical to the first review.
- **Finding 2 (26.04 not covered by in-repo CI) — unchanged, non-blocking, acknowledged.** Still true; out of scope per the author's plan to add Resolute to the Launchpad recipe after merge.

### Independent verification performed this pass

- Diff is exactly the same 2-line change as before: `debian/control` build-dep alternative, `meson.build` `cpp_std=c++14` → `c++17`. Nothing else touched — proportionate to the stated goal.
- Grepped all 9 `tests/capi/*.cc` files (the only C++ sources in the repo besides the Android JNI glue) for constructs removed/restricted in C++17 (`auto_ptr`, `random_shuffle`, `unary_function`/`binary_function`, `bind1st`/`bind2nd`/`ptr_fun`, dynamic exception specs, `register` as a storage-class keyword). No real hits — only benign uses of the English word "register" in comments.
- Repo-wide grep for `cpp_std`, `c++14`, `c++11`, `gnu++` found only the two expected sites: `meson.build` (now `c++17`) and `packaging/machine-learning-api.spec` (strips `-std=gnu++11` from `CXXFLAGS` so meson's `cpp_std` wins on Tizen GBS too). No other subproject or build file pins a conflicting standard.
- `java/android/nnstreamer/src/main/jni/Application.mk` already hardcodes `APP_CPPFLAGS := -std=c++17`, so the NDK build is unaffected and was already ahead of this change.
- `meson_version: '>=0.50.0'` in `meson.build` is compatible with `cpp_std=c++17`.
- `.github/workflows/pdebuild.yml` only builds `ubuntu-22.04` (jammy) with `gcc-9`-resolvable toolchain; it runs on every PR and will build the 9 gtest binaries under the new `cpp_std=c++17`, so any real C++17 incompatibility introduced in this or a future PR would fail that CI job before merge. There is no 24.04 or 26.04 job in-repo (matches the PR's own statement and its follow-up plan), so 26.04 verification stays manual until the Launchpad recipe change lands.
- No architecture or public API change; no documentation update needed.
- Build-script-only change with no functional/behavioral change to installed libraries — the PR body's stated rationale for adding no new test case holds up under inspection (existing gtest suite is what re-validates the C++17 bump; the fallback compiler line has no test surface of its own beyond "the package still builds").

### Note

The PR is still marked `DRAFT` with a `DO NOT MERGE` label — presumably intentional pending the Launchpad recipe follow-up mentioned in the PR body, not a code issue.

Nothing else to add; this remains the minimal change needed and continues to mirror the fallback pattern already merged in nnstreamer and nnstreamer-edge.

myungjoo:
Libraries we use are requiring C++17 these days :(

